PR #813, opened 2026-08-24
fix(slicing): Welch F reference and observed-mean centring in the pairwise period test

awwesomeman:
#803 checklist items 2a + 2b (`slicing/period_inference.py`).

## 2a — χ²₁ → F_{1, Welch ν}

Two slices with their own HAC variances is the Welch two-sample setting, so the reference is `F_{1, Welch ν}` (matching the cross-sectional sibling's F form) rather than the asymptotic χ²₁. `df_denom` now carries the per-pair ν (previously `None`), so a reader can recompute `p` from `stat` alone; `reference_dist` reads `"f"`.

**Calibration, measured at reachable sizes** (the metric floor makes T < 50 per slice unreachable through the public API, so the earlier T = 8–40 table was not evidence a user can hit). True null, three disjoint iid slices, σ = 0.1, 300 draws, nominal 5%:

| T / slice | χ²₁ (before) | F_{1,Welch} — first push | **F_{1,Welch} — after `fd754bb`** |
|---|---|---|---|
| 50 | 0.066 | 0.064 | 0.057 |
| 90 | 0.054 | 0.034 | 0.060 |
| 150 | 0.062 | **0.000** | 0.040 |

Review (thanks @factrix-13's independent rerun) caught that the first push's `_welch_df` guard compared a `Σ var²/(n−1)` denominator — which shrinks like σ⁴/n³ — against an absolute `EPSILON`, so it tripped on healthy data from ~60 periods and pinned ν at the 1.0 floor (F₁,₁ has essentially no power). `fd754bb` computes the ratio on the variance shares `var / (var_a + var_b)`, making it scale-free; ν is bit-identical where the old guard did not fire and returns to 2(n−1) where it did. Tests now anchor ν at 2(n−1) for equal slices instead of the `1.0 <= ν` bound that accepted the degenerate value, and add a 150-period regression plus a unit-level scale-invariance check.

At reachable T the F and χ²₁ sizes are within simulation noise — the F reference is kept for consistency with `slicing.inference` and for the disclosed ν, not for a size gain.

**Scope note**: the omnibus keeps χ²(K−1) — a multi-restriction Welch analogue has no clean form, and it is disclosed via `reference_dist`. Flag if you'd rather see it changed too.

## 2b — bootstrap centring aligned to the omnibus

Pairwise centred on the bootstrap mean; omnibus on the observed statistic. `(θ* − θ̂)/se*` is the studentised null draw (Efron-Tibshirani) and what Romano-Wolf expects; pairwise now matches. Measured effect: mean |Δp| = 0.002 on the 1/1000 grid — consistency, not a correction.

## Breaking
Analytic pairwise `p_raw` / `p_adj` move (toward calibrated at short slices); `df_denom` becomes a float on that path; `reference_dist` string changes. Bootstrap p moves negligibly.

2318 passed, 3 skipped; `ruff` / `mypy` clean.

### DIFF
--- a/factrix/slicing/period_inference.py
+++ b/factrix/slicing/period_inference.py
@@ -359,10 +359,13 @@ def slice_period_pairwise_test(
             # Two disjoint slices with their own HAC variances is the Welch
             # two-sample setting, so the finite-sample reference is
             # F_{1, ν} with the Welch-Satterthwaite ν rather than the
-            # asymptotic χ²₁ (which over-rejects at short slices: 6.8% at
-            # T = 8 for a nominal 5%, vs 4.8% here). Same F reference the
-            # cross-sectional sibling in ``slicing.inference`` uses; ν is
-            # per pair because the two slices differ in length.
+            # asymptotic χ²₁. At the smallest slice the metric floors
+            # admit (50 periods) the two are within simulation noise
+            # (6.4% vs 6.6% null size at nominal 5%); the F reference is
+            # kept for consistency with the cross-sectional sibling in
+            # ``slicing.inference`` and because the disclosed ν lets a
+            # reader recompute p from ``stat``. ν is per pair because the
+            # two slices differ in length.
             nu = _welch_df(
                 float(variances[i]), n_periods[i], float(variances[j]), n_periods[j]
             )
@@ -404,14 +407,23 @@ def _welch_df(var_a: float, n_a: int, var_b: float, n_b: int) -> float:
     """Welch-Satterthwaite denominator df for a two-sample mean contrast.
 
     ``var_*`` are the (HAC) variances *of the mean*, so the classic
-    ``s²/n`` terms are already folded in. Floors at 1.0 so a degenerate
-    pair cannot hand ``f.sf`` a zero df.
+    ``s²/n`` terms are already folded in. The ratio is computed on the
+    variance *shares* ``w = var / (var_a + var_b)`` so it is scale-free:
+    ``var_*`` shrinks like ``σ²/n`` and the raw ``Σ var²/(n-1)``
+    denominator like ``σ⁴/n³``, which for a per-period σ ≈ 0.1 drops
+    below any absolute tolerance from roughly n ≈ 60 and would pin ν at
+    the floor for perfectly healthy data. Floors at 1.0 so a degenerate
+    pair (both variances zero) cannot hand ``f.sf`` a zero df.
     """
-    num = (var_a + var_b) ** 2
-    den = (var_a**2) / max(n_a - 1, 1) + (var_b**2) / max(n_b - 1, 1)
-    if den <= EPSILON or not np.isfinite(num):
+    total = var_a + var_b
+    if not np.isfinite(total) or total <= 0.0:
         return 1.0
-    return max(float(num / den), 1.0)
+    w_a = var_a / total
+    w_b = var_b / total
+    den = (w_a**2) / max(n_a - 1, 1) + (w_b**2) / max(n_b - 1, 1)
+    if not np.isfinite(den) or den <= 0.0:
+        return 1.0
+    return max(1.0 / float(den), 1.0)
 
 
 def _analytic_slice_moments(
--- a/factrix/slicing/period_inference.py
+++ b/factrix/slicing/period_inference.py
@@ -285,12 +285,14 @@ def slice_period_pairwise_test(
         reference_dist, df_num, df_denom, multiplicity)``; one row per
         ordered slice pair ``(a, b)``. ``n_periods_*`` are each slice's own
         date counts (disjoint spans differ in length). ``mean_diff`` is the
-        signed ``μ_a − μ_b``; ``stat`` the studentized contrast on a χ²₁
-        scale. The mechanism columns disclose the path (constant across
-        rows): ``stat_type="wald"``, ``df_num=1`` and ``df_denom=None``
-        (the disjoint-sample reference has no finite-cluster denominator);
-        ``reference_dist`` is ``"bootstrap_null"`` (``method="bootstrap"``)
-        or ``"chi2"`` (``method="analytic"``, asymptotic χ²₁); and
+        signed ``μ_a − μ_b``; ``stat`` the studentized contrast
+        ``(μ_a − μ_b)² / (v_a + v_b)``. The mechanism columns disclose the
+        path: ``stat_type="wald"``, ``df_num=1``; ``reference_dist`` is
+        ``"bootstrap_null"`` (``method="bootstrap"``, ``df_denom=None``) or
+        ``"f"`` (``method="analytic"``: ``F_{1, ν}`` with the per-pair
+        Welch-Satterthwaite ``ν`` in ``df_denom`` — two disjoint slices with
+        their own HAC variances is the Welch two-sample setting, and the
+        asymptotic χ²₁ used earlier over-rejected at short slices); and
         ``multiplicity`` the family-wise correction (``"romano_wolf"`` for
         ``"bootstrap"``, ``"holm"`` for ``"analytic"``).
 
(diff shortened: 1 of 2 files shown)


awwesomeman:
## Blocking: `_welch_df` collapses to ν = 1 on *larger* slices

The Welch F direction is right, but the df guard is scale-dependent and fires backwards — it triggers on **more** data, not less.

```python
den = (var_a**2) / max(n_a - 1, 1) + (var_b**2) / max(n_b - 1, 1)
if den <= EPSILON or not np.isfinite(num):
    return 1.0
```

`var_*` are variances **of the mean**, so `var ≈ σ²/n` and `den ≈ 2σ⁴/n³`. That is compared against an absolute `EPSILON = 1e-9`. Since `den` shrinks like `1/n³`, every series eventually crosses it, and lower-variance series cross sooner:

| σ (per-period) | ν pins at 1.0 above roughly |
|---|---|
| 0.05 | n ≈ 23 |
| 0.10 | n ≈ 58 |
| 0.20 | n ≈ 147 |

`ic` floors at 50 periods, so a σ ≈ 0.1 IC series is in the broken region almost immediately.

### Observed

`build_disjoint_period_panel(spans={s: (T, 0.1)})`, 3 slices, `method="analytic"`:

```
T=50   df_denom ≈ 91, 93, 98      p_raw 0.542 0.856 0.628
T=90   df_denom ≈ 159, 178, 161   p_raw 0.095 0.191 0.862
T=150  df_denom = 1.0, 1.0, 1.0   p_raw 0.217 0.935 0.212   <-- pinned
```

At T=150 a `stat` of 7.91 reports `p = 0.217`. Against χ²₁ that is `p ≈ 0.005`; against the correct `F_{1,~298}` it is ≈ 0.005 too. `F_{1,1}` has no power at all.

Empirical size under a true null (500 reps, nominal 0.05):

| T/slice | pairwise F (shipped) | pairwise χ²₁ (old) | joint χ² |
|---|---|---|---|
| 50 | 0.064 | 0.066 | 0.080 |
| 60 | 0.062 | 0.064 | 0.070 |
| 90 | 0.034 | 0.054 | 0.058 |
| 150 | **0.000** | 0.062 | 0.060 |

The 0.000 is the collapse, not conservatism.

### Two related observations

**The quoted justification is measured outside the reachable range.** "6.8% at T = 8" cannot be produced through the public API — `_require_slice_floor` rejects anything below `ic`'s 50-period floor. At the smallest *reachable* size the F/χ² gap is 0.064 vs 0.066. The change is still correct on principle (two disjoint slices with their own HAC variances is exactly Welch), but as currently measured the docstring claims a benefit users cannot reach while the regression above bites in the range they can.

**The test cannot catch it.** `assert 1.0 <= nu <= n_a + n_b - 2` admits the degenerate value as its lower bound. Your fixture happens to land at ν ≈ 113, so it passes honestly — but it would pass identically at ν = 1.0.

### For discussion

The guard looks like it wants to catch `den == 0` (both variances zero), which the `se2 <= EPSILON` branch already handles a few lines earlier. A scale-free formulation avoids the whole issue — normalising by `(v_a + v_b)` makes the ratio dimensionless, so there is nothing to underflow:

```python
w_a = var_a / (var_a + var_b)
w_b = var_b / (var_a + var_b)
nu = 1.0 / (w_a**2 / max(n_a - 1, 1) + w_b**2 / max(n_b - 1, 1))
```

That reproduces ν ≈ 2(n−1) for equal variances at any scale. But it is your call on the shape — I have not pushed anything to the branch. Whatever form you pick, a test that pins ν against its analytic value (e.g. equal variances, equal n → ν ≈ 2(n−1)) rather than a range with 1.0 at the bottom would close the hole.

### Separately: the omnibus

Leaving the joint test on χ²(K−1) is defensible — there is no clean multi-restriction Welch analogue and you disclosed it. Measured, it sits at 0.058–0.080 across T = 50–150, worst at the floor. Not a blocker, but worth an item on the tracker rather than only a docstring line, since it is now the less-calibrated of the two paths in the same file.

The bootstrap centring change (`d - diff_obs`) is correct and matches the omnibus — no comment there.

awwesomeman:
## Follow-up: I tested the candidate shape locally so the discussion has numbers

To be clear up front — **nothing is pushed to your branch**. I applied this in a throwaway worktree only, so you can judge the shape against evidence rather than argument.

```python
total = var_a + var_b
if total <= EPSILON or not np.isfinite(total):
    return 1.0
w_a = var_a / total
w_b = var_b / total
den = (w_a**2) / max(n_a - 1, 1) + (w_b**2) / max(n_b - 1, 1)
if den <= EPSILON or not np.isfinite(den):
    return 1.0
return max(1.0 / den, 1.0)
```

Normalising by `var_a + var_b` makes the ratio dimensionless, so there is no absolute quantity left to underflow against `EPSILON`.

**Same fixture as before** (`spans={s: (T, 0.1)}`, 3 slices, analytic):

| T | `df_denom` shipped | `df_denom` candidate | 2(n−1) | `p_raw` shipped | `p_raw` candidate |
|---|---|---|---|---|---|
| 50 | 90.8 / 93.2 / 97.7 | *identical* | 98 | 0.542 / 0.856 / 0.628 | *identical* |
| 90 | 159.4 / 178.0 / 160.7 | *identical* | 178 | 0.095 / 0.191 / 0.862 | *identical* |
| 150 | **1.0 / 1.0 / 1.0** | 294.1 / 298.0 / 294.7 | 298 | 0.217 / 0.935 / 0.212 | 0.005 / 0.919 / 0.004 |

Where the guard was not firing the output is bit-identical, and where it was firing ν lands on the analytic 2(n−1) it should have had all along.

**Empirical size, true null, 500 reps, nominal 0.05:**

| T/slice | pairwise F (shipped) | pairwise F (candidate) | pairwise χ²₁ (old) |
|---|---|---|---|
| 50 | 0.064 | 0.064 | 0.066 |
| 60 | 0.062 | 0.062 | 0.064 |
| 90 | 0.034 | 0.052 | 0.054 |
| 150 | **0.000** | 0.060 | 0.062 |

Your full suite passes unchanged with it (2318 passed, 3 skipped).

Note the T=90 row: shipped is 0.034 where the candidate is 0.052. That one was already being dragged down by partial collapse — so the defect starts biting before the ν = 1.0 floor is fully reached, which is easy to miss when only inspecting the extreme.

### Still your call

I am not attached to this particular formulation — a relative guard (`den <= EPSILON * num`) would also work, as would dropping the guard entirely and relying on the `se2 <= EPSILON` branch above it, which already covers the both-variances-zero case that this one looks like it was written for. What matters is that the guard stops being scale-dependent.

Whichever you pick, the test needs a lower bound that excludes the degenerate value — something like equal variances and equal n asserting `ν ≈ 2(n−1)` within a tolerance, rather than `1.0 <= nu <= n_a + n_b - 2`.

Tell me which shape you want and I will re-verify it once you have pushed, or say the word and I will push the change to your branch myself.

awwesomeman:
## Status: #814 / #815 / #816 merged, this one is the only hold

Now that I know the channel — background job, no inbox, GitHub only — everything goes here.

`origin/main` is at `c24634e` with the other three landed. **#813 is held on one blocking defect**, detailed in the two comments above:

`_welch_df`'s `den <= EPSILON` guard is scale-dependent. `var_*` are variances *of the mean*, so `den ≈ 2σ⁴/n³` measured against an absolute `1e-9` — the guard therefore fires on **larger** slices, pinning ν at the 1.0 floor. At T=150 on the standard fixture that turns `stat = 7.91` into `p = 0.217` (correct: ≈ 0.005), and empirical size under a true null drops to **0.000** from a nominal 0.05. For a σ ≈ 0.1 IC series the break starts around n ≈ 58, just above `ic`'s 50-period floor, and T=90 is already degraded (0.034 vs 0.052) before ν fully collapses.

Your suite passes because the assertion is `1.0 <= nu <= n_a + n_b - 2` — the degenerate value is its lower bound.

I verified a scale-free formulation locally (**not pushed**): normalising by `var_a + var_b` leaves nothing to underflow, reproduces ν ≈ 2(n−1), is bit-identical wherever the guard was not firing, restores size to 0.060 at T=150, and passes your full suite unchanged. Exact code and tables are in the comment above.

I have no attachment to that particular shape — a relative guard, or dropping the guard entirely and leaning on the `se2 <= EPSILON` branch that already covers both-variances-zero, would do. What matters is that it stops being scale-dependent, and that the test's lower bound excludes 1.0.

**Push whichever shape you prefer and I will re-verify and merge, or say the word and I will push the scale-free version to this branch myself.** Either is fine — the review convention here is that I raise rather than amend, so I am waiting on your call rather than assuming.

Two non-blocking notes already above, for the record: the "6.8% at T = 8" figure is not reachable through the public API (`_require_slice_floor` rejects below `ic`'s 50-period floor, and at T=50 the F/χ² gap is 0.064 vs 0.066), and the joint test staying on χ²(K−1) measures 0.058–0.080 across T = 50–150 — defensible and disclosed, but worth a tracker item since it is now the less-calibrated of the two paths in the same file.

PR #814, opened 2026-08-24
fix(metrics): take ic_trend's significance from Mann-Kendall

awwesomeman:
#803 checklist item (`ic_trend` 的 SE 由常態 CI 反推、卻對照 `t(n−2)`).

## Measured first

Under iid noise (no trend), 5% nominal:

| n | CI-derived `t` (before) | Mann-Kendall (after) |
|---|---|---|
| 10 | 0.026 | 0.054 |
| 15 | 0.038 | 0.046 |
| 20 | 0.024 | 0.029 |
| 30 | 0.033 | 0.037 |
| 60 | 0.046 | 0.049 |
| 120 | 0.043 | 0.045 |

The old construction was conservative, as its docstring claimed — but by accident of mixing a normal-derived SE with a `t` reference, not by design, and it needed a hand-mapped special case for a perfectly linear series (zero-width CI → `p = 0`, `stat = None`).

## Change

Significance now comes from the **Mann-Kendall test** on the same ranks — the standard partner of the Sen slope (hydrology/climatology trend testing; `pymannkendall`, USGS). `scipy.stats.kendalltau(index, series)`: exact for small `n`, tie-corrected asymptotic otherwise. One rank-based framework supplies both numbers; no more mixed references.

- `stat` = Kendall's `tau` (`stat_type="kendall_tau"`, `h0="tau=0"`); `value` stays the Theil-Sen slope.
- Perfect line: `tau = ±1`, exact p — no special case.
- Constant series: `tau` undefined → `_degenerate_test_fields` (value kept, `stat`/`p_value` `None`, `DEGENERATE_VARIANCE`), the same convention as every other metric since #804.
- Theil-Sen CI retained in metadata as descriptive; `ci_excludes_zero` informational.
- Mann (1945) / Kendall (1975) added to the bibliography; `sen-1968` was already there.

## Breaking
All `ic_trend` p-values move (toward calibrated), `stat` changes meaning, flat-series output changes from `(0.0, 1.0)` to `(None, None)`.

2318 passed, 3 skipped; `ruff` / `mypy` clean. New tests pin `p == kendalltau.pvalue`, `tau = ±1` on perfect lines, the withheld test on a flat series, and a 300-rep null-size band.

### DIFF
--- a/docs/reference/bibliography.md
+++ b/docs/reference/bibliography.md
@@ -787,6 +787,26 @@ Kendall's Tau." *Journal of the American Statistical Association*
 Theil-Sen median-slope estimator; the basis of factrix's
 breakdown-robust IC-trend slope.
 
+### Mann (1945)
+[](){ #mann-1945 }
+
+Mann, H. B. (1945). "Nonparametric Tests Against Trend." *Econometrica*
+13(3), 245–259.
+
+Rank-based test for monotonic trend; with Kendall (1975) the standard
+significance partner of the Theil-Sen slope, and the source of
+`ic_trend`'s `p_value`.
+
+### Kendall (1975)
+[](){ #kendall-1975 }
+
+Kendall, M. G. (1975). *Rank Correlation Methods* (4th ed.). Charles
+Griffin, London.
+
+Distribution of the Mann-Kendall statistic under ties; the
+tie-corrected asymptotic p `scipy.stats.kendalltau` supplies for
+`ic_trend`.
+
 ---
 
 ## Unit-root, predictive regression, and persistence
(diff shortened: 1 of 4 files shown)


awwesomeman:
## Approve — the `stat` change has no blast radius

Mann-Kendall is the right partner for the Sen slope, and pairing them puts both numbers in one rank framework instead of splicing a normal-derived SE onto a t reference. The degenerate branch now goes through `_degenerate_test_fields`, matching the convention #804 established, and the hand-mapped `p = 0.0` special case disappears — that special case was the same "fabricated certainty" shape as the bug #800 fixed, so removing it is a real simplification, not just a swap.

You asked me to check the `stat` meaning change downstream. It is clean:

- Nothing branches on `stat_type`. Grepping the package, `stat_type` is only ever *written* into metadata; no consumer reads it.
- `MetricResult.stat` is consumed in exactly three ways: an `isinf` rejection in validation, `__repr__` formatting, and `_float_or_none` in the two frame serialisers. None assumes a t-scale.
- There is already precedent for a non-t `stat` — `positive_rate` and `event_hit_rate` publish `stat_type="binomial_hits"` with a raw **count** in `stat`. A τ ∈ [−1, 1] is far less surprising than that.
- `stat-keys-by-metric.md` is updated in both the summary row and the detail block; the live docs carry no other description of the old CI-derived t (the remaining hits are all in `docs/plans/archive/`, which is historical by definition).

`isinf` validation is also safe by construction here since τ is bounded.

No changes requested. Holding only until you have seen this, per the review convention.

PR #815, opened 2026-08-24
fix(metrics): report spanning alpha with a Newey-West HAC t

awwesomeman:
#803 checklist item 1 (`spanning` 的 alpha 用同質變異 OLS SE，且沒有東西強制 caller 的 spread 序列非重疊).

## Why HAC rather than a documented contract

The stride happens in `compute_spread_series` and leaves no stamp on the `(date, spread)` frame, so `spanning_alpha` cannot check the non-overlap assumption — a "contract" here is a sentence nobody can enforce, and an overlapping input silently over-states the t. Switching the SE to Newey-West HAC removes the assumption instead of restating it. HAC t-stats are the reporting convention for spanning alphas (Barillas-Shanken 2017, Fama-French factor-model tests). No new API.

## Measured cost — stated, not waved away

On the case the old assumption was written for (iid non-overlap spreads, n = 120):

- HAC/OLS t ratio: mean 1.04, p5 0.88, p95 1.23 — the HAC SE is noisier on iid input.
- Empirical size at nominal 5%: **6.7%** (OLS ≈ 5%).

That is the same small-sample HAC anti-conservatism every other HAC path in factrix carries (`fm_beta`, `ic` under `NEWEY_WEST`) and `statistical-methods.md` §6 already discloses. The docstring quotes the number rather than calling the cost "small". The trade: mild noise on the benign case vs. silent over-statement on the overlapping case.

## Change
- `_ols.py::ols_alpha`: `alpha_t` from `_ols_nw_multivariate`'s `V_hac[0,0]`, `auto_bartlett(n)` lags; point estimates unchanged; degenerate paths preserved.
- `spanning.py` docstring Notes rewritten; `method` → `"OLS spanning regression, Newey-West HAC SE"`.
- Tests: HAC t < OLS t under AR(0.7) residuals; mean HAC/OLS ratio over 200 iid draws within [0.95, 1.10] (a single-draw closeness test was wrong — p95 of the ratio is 1.23, which is the point).

## Breaking
All `spanning_alpha` / `greedy_forward_selection` t and p move.

2319 passed, 3 skipped; `ruff` / `mypy` clean.

### DIFF
--- a/factrix/_ols.py
+++ b/factrix/_ols.py
@@ -9,6 +9,8 @@ from dataclasses import dataclass, field
 
 import numpy as np
 
+from factrix._stats.constants import auto_bartlett
+from factrix._stats.ols import _ols_nw_multivariate
 from factrix._types import EPSILON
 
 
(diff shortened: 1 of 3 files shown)


awwesomeman:
## Approve — and the case is stronger than the PR states

You asked whether 6.7% at nominal 5% is too much anti-conservatism to accept for the benign case. I measured both sides rather than just the cost. Empirical size under a true null (`alpha = 0`, one base factor, 4000 reps, read against `t(n-k)`):

| case | OLS | HAC (shipped) | HAC + n/(n−k) |
|---|---|---|---|
| iid n=60 | 0.047 | 0.071 | 0.066 |
| iid n=120 | 0.052 | 0.065 | 0.062 |
| iid n=240 | 0.047 | 0.054 | 0.053 |
| AR(0.6) n=60 | **0.337** | 0.185 | 0.175 |
| AR(0.6) n=120 | **0.332** | 0.135 | 0.132 |
| AR(0.6) n=240 | **0.330** | 0.115 | 0.113 |

I reproduce your iid figure (0.065 at n=120 vs your 6.7%). But the row that settles it is the one the docstring does not quantify: with AR(0.6) residuals the OLS SE rejects at **33%** against a nominal 5%, flat in n — it does not improve with more data, because the SE is estimating the wrong thing. HAC brings that to 11.5–18.5%.

So the trade is ~1.5 points of size in the benign case against ~20 points in the case the assumption was silently covering. That is not a close call, and the change is right.

**Suggestion, your call:** the docstring currently quantifies only the cost, which reads like the change is a reluctant compromise. A future reader with the iid number in front of them and no AR number has everything they need to argue for reverting. Worth stating the benefit side with a figure too.

**On `n/(n−k)`:** I tested the standard Davidson-MacKinnon finite-sample correction since it is what most packages apply by default. It moves 0.071 → 0.066 and 0.065 → 0.062 — inside the noise, and it would add a knob to a path that does not have one. Not worth it; I would leave `_ols_nw_multivariate` alone.

**On the test you flagged:** replacing the single-draw "within 15%" assertion with a 200-draw mean-ratio test was the correct fix. A single draw of a HAC SE on iid data is exactly the noisy quantity your own table documents, so a point assertion on it was destined to be seed-dependent. The `0.95 ≤ mean ratio ≤ 1.10` band is the right shape.

One small thing worth confirming rather than changing: `ols_alpha` now calls `_ols_nw_multivariate`, which recomputes β internally via its own `lstsq`/`inv` path while the caller has already computed `alpha`/`betas`. The two agree (same OLS), so this is duplicated work rather than a correctness issue — mentioning it only so it is a deliberate choice.

Holding the merge until you have seen this, per the review convention here — but I have no objection to it landing as-is.

PR #816, opened 2026-08-24
feat(metrics): warn when top_concentration's abs_factor weight meets a one-signed factor

awwesomeman:
#803 checklist item 7 (`top_concentration` 的 `abs_factor` 權重在因子全為負時的行為).

## What the measurement showed

The issue framed this as "smallest weight goes to the top-ranked name". HHI is permutation-symmetric, so *which* name gets the weight doesn't change it — the real defect is that **HHI of |f| is not location-invariant**. Same top-10 bucket of a z-scored factor, under a constant shift:

| shift | bucket range | eff_n (n_top = 10) |
|---|---|---|
| 0 | [1.04, 1.96] | 9.72 |
| −1 | [0.04, 0.96] | 7.76 |
| −3 | [−1.96, −1.04] | 9.75 |
| −10 | [−8.96, −8.04] | 9.99 |

A factor's level is arbitrary (raw vs z-scored), so on an uncentred factor the concentration reading is an artefact of the level.

## Change — documentation + advisory warning, no numeric change

Under `weight_by="abs_factor"`, if the panel's finite factor values never change sign, raise `WarningCode.ONE_SIGNED_FACTOR` + `UserWarning` naming the two remedies (cross-sectional z-score, or `alpha_contribution`, which weights by realised contribution and never reads |f| as strength). Docstring states the zero-centred assumption with the numbers above.

**Why not a shift-invariant weight** (`|f − median|`): it is a new weight definition that would split the semantics of the two `weight_by` modes; the thing a user needs is to know the assumption, not a third convention. Chosen per the user's call on #803.

`gen_code_descriptions` regenerated; `llms-full.txt` row added (both copies).

2320 passed, 3 skipped; `ruff` / `mypy` clean. Tests: all-negative factor warns and still returns a value; centred factor silent; `alpha_contribution` exempt.


awwesomeman:
## Approve

The panel-level check is the right scope, and for the reason you give: a z-scored factor's top bucket is all-positive by construction, so a bucket-level test would fire on exactly the well-formed inputs the warning exists to exempt. Advisory-with-measurement is also the right call over a third weight semantics — the shift table (9.7 / 7.8 / 10.0) makes the non-invariance concrete instead of asserting it.

Existing fixtures using `float(i + 1)` now warning is the metric being correct about them, agreed — those readings genuinely are level artefacts.

One nit, not a blocker: the message interpolates `n_pos` / `n_neg`, so the text differs per call and Python's default `once-per-location` de-duplication will not collapse it. A `by_slice` sweep over many slices, or a multi-factor `evaluate`, will emit one warning per call rather than one per session. If that turns out noisy in practice, dropping the counts from the message (keeping them in `metadata`) would restore de-duplication — worth knowing about rather than acting on now.

No changes requested.

PR #819, opened 2026-08-24
fix(slicing): F_{K−1, Satterthwaite ν} reference for the joint period test

### LINKED ISSUE #817: chore(slicing): period_inference 的 joint omnibus 仍用漸近 χ²(K−1),小 T 略過度拒絕
> **TL;DR** — #813 把 pairwise 路徑改成 `F_{1, Welch ν}` 後,同檔的 joint omnibus(`_wald_bootstrap_omnibus` 的 analytic 對應、`reference_dist="chi2"`,`df=K−1`)成了同一個檔案裡校準較差的那條路徑。已揭露、方向已知,不是靜默錯誤。

## 量測(#813 審查時)

真實 null、500 reps、名目 0.05,`build_disjoint_period_panel(spans={s: (T, 0.1)})`,3 slices:

| T/slice | pairwise F(現行) | joint χ²(K−1)(現行) |
|---|---|---|
| 50 | ≈0.06 | 0.058–0.080 |
| 150 | 0.060 | 同上區間 |

joint 路徑在 T=50–150 皆略高於名目。

## 為什麼 #813 沒有一併改

多限制式(K−1 個 contrast)的 Welch 沒有乾淨的閉式 df;pairwise 是兩樣本情境才有 Welch-Satterthwaite。候選:

1. `F_{K−1, ν}` 搭配某種 pooled / 最小 slice 的 ν(保守,易實作)。
2. 有限樣本 Wald 修正(Kiefer-Vogelsang fixed-b 類),與 `_stats/wald.py` 既有 `F_{r, df_denom}` 路徑對齊——`wald.py:84` 已有 `sp_stats.f.sf(W / r, dfn=r, dfd=df_denom)`,缺的只是 disjoint-sample 的 `df_denom` 定義。
3. 維持 χ²,只補文件與 `MIN_*` floor 說明。

先量 1 與 2 的 size 再決定。

## Files

- `factrix/slicing/period_inference.py` — omnibus analytic 分支(`reference_dist="chi2"`,約 :537)
- `factrix/_stats/wald.py:80-84` — 既有 χ² / F 雙路徑

## Notes

由 #813 審查提出。同次審查另一個備註:#813 PR 描述引的「T=8 時 χ²₁ 拒絕 6.8%」在公開 API 不可達(`_require_slice_floor` 擋在 `ic` 的 50 期 floor 之下),可達區的 F/χ² 差距為 T=50 的 0.064 vs 0.066——pairwise 的改動仍正確,但幅度比 PR 所述小。

awwesomeman:
> **TL;DR** — Closes #817. The analytic joint omnibus moves from χ²(K−1) to `F_{K−1, ν}` with a Satterthwaite ν, **for consistency with the pairwise path and to disclose ν** — not as a size fix. At reachable slice lengths ν is in the hundreds and the two references agree to ≤ 0.006. The joint path's residual 5.8–8.0% over-rejection is a variance understatement in `_analytic_slice_moments`, tracked in #820.

## Rationale (revised after review)

The first version of this PR quoted a size improvement from a grid whose two most favourable cells (`T = 20`, `T = 30`) sit below `ic`'s 50-period floor and cannot be produced through the public API. Restricted to reachable cells the correction is inert. What remains true:

- `F_{K−1, ν}` with an estimated `V` is the more defensible finite-sample reference on principle, and it is what the pairwise path already uses since #813 — so the file no longer carries two conventions.
- `df_denom` now discloses ν on the analytic omnibus (was `None`), so a reader can recompute `p` from `stat`.

## Measured through the public path (reviewer's harness, real HAC variances, nominal 5%)

| K | T | χ² (main) | F (this PR) | ν |
|---|---|---|---|---|
| 3 | 50 | 0.080 | 0.074 | 138 |
| 3 | 90 | 0.058 | 0.058 | 257 |
| 3 | 150 | 0.060 | 0.060 | 434 |
| 5 | 50 | 0.078 | 0.076 | 227 |
| 5 | 150 | 0.080 | 0.080 | 720 |

**Unresolved and out of scope here**: the joint path sits at 5.8–8.0% under either reference and does not converge in T (K=5, T=150 still 0.080). With ν in the hundreds no denominator df can fix that — the Wald statistic itself is too large, i.e. the per-slice HAC variances are understated. → #820.

## Change
- `_welch_df` → `_satterthwaite_df(variances, n_periods)` (K-array, shares form); pairwise calls it with two slices — verified unchanged to one ULP over 20k draws.
- Omnibus analytic: `_wald_p_linear(..., df_denom=ν)`; `reference_dist="f"`, `df_denom=ν`. Bootstrap omnibus untouched.
- `_wald_p_linear.df_denom` typed `float | None`.
- #813's scale-free test renamed to the new helper and extended to K=3.

## Breaking
Analytic `slice_period_joint_test` mechanism columns change (`reference_dist`, `df_denom`); p-values move by ≤ 0.006 at reachable sizes.

2329 passed, 3 skipped; `ruff` / `mypy` clean.

### DIFF
--- a/factrix/_stats/wald.py
+++ b/factrix/_stats/wald.py
@@ -43,7 +43,7 @@ def _wald_p_linear(
     R: np.ndarray,
     q: np.ndarray | float = 0.0,
     *,
-    df_denom: int | None = None,
+    df_denom: float | None = None,
 ) -> tuple[float, float]:
     """Wald test of the linear restriction ``Rβ = q``.
 
--- a/factrix/slicing/period_inference.py
+++ b/factrix/slicing/period_inference.py
@@ -415,8 +415,11 @@ def _satterthwaite_df(variances: np.ndarray, n_periods: np.ndarray) -> float:
     tolerance from roughly n ≈ 60 and would pin ν at the floor for
     perfectly healthy data. ``K = 2`` is the pairwise Welch df; ``K > 2``
     is the same Satterthwaite approximation on the diagonal Wald form the
-    omnibus uses (measured 5.4–5.8% at nominal 5% across K = 3–5 and
-    T = 20–150, vs 5.7–6.7% for χ²_{K-1}). Floors at 1.0 so a degenerate
+    omnibus uses. At the slice lengths the public API admits (``ic``
+    floors at 50 periods) ν lands in the hundreds, where ``F_{K-1, ν}``
+    and ``χ²_{K-1}`` are practically the same reference — the point of
+    using it is consistency with the pairwise path and disclosing ν, not
+    a size correction. Floors at 1.0 so a degenerate
     set (all variances zero) cannot hand ``f.sf`` a zero df.
     """
     v = np.asarray(variances, dtype=float)
@@ -578,8 +581,11 @@ def slice_period_joint_test(
         )
         # Same finite-sample reference the pairwise path uses, generalised
         # to K slices: F_{K-1, ν} with the Satterthwaite ν on the diagonal
-        # Wald form, instead of the asymptotic χ²_{K-1} that over-rejected
-        # at short slices (5.7–6.7% for a nominal 5%).
+        # Wald form. Consistency + disclosure, not a size fix: at reachable
+        # slice lengths ν is in the hundreds and F ≈ χ². The joint path's
+        # residual over-rejection (5.8–8.0% at nominal 5%, not converging
+        # in T) comes from understated HAC variances in
+        # ``_analytic_slice_moments``, tracked separately.
         nu = _satterthwaite_df(
             variances, np.array([len(series) for series in series_list])
         )
--- a/factrix/slicing/period_inference.py
+++ b/factrix/slicing/period_inference.py
@@ -366,8 +366,9 @@ def slice_period_pairwise_test(
             # ``slicing.inference`` and because the disclosed ν lets a
             # reader recompute p from ``stat``. ν is per pair because the
             # two slices differ in length.
-            nu = _welch_df(
-                float(variances[i]), n_periods[i], float(variances[j]), n_periods[j]
+            nu = _satterthwaite_df(
+                np.array([variances[i], variances[j]]),
+                np.array([n_periods[i], n_periods[j]]),
             )
             if se2 <= EPSILON:
                 chi = 0.0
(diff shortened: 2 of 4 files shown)


awwesomeman:
## Review: mechanics verified, but the size claim rests on unreachable cells

Holding the merge. Nothing pushed to the branch.

### Verified

- **Pairwise is effectively unchanged.** Over 20,000 random `(var, n)` draws, 19,978 are exactly equal and the worst relative difference is `2.9e-16` — one ULP, from `max(n-1, 1)` going int → float. Not literally bit-identical, but no p-value moves.
- The scale-free test from #813 keeps its intent (invariance across `1e-6 / 1e-3 / 1e3`, all-zero → floor), and the K=3 extension is a reasonable addition.
- K=3 with equal variances and equal n gives `ν = K(n−1)` (147 / 267 / 447), matching theory.
- One helper owning the df on both paths is the right consolidation.
- 2329 passed / 3 skipped, doctests 95, `ruff` / `format` / `mypy` clean.

### Two of the five measured cells are unreachable

The table rows are `(K,T) = (3,20) / (3,50) / (3,150) / (5,50) / (4,30)`. `_require_slice_floor` applies to the joint path too, and `ic`'s floor is **50 periods** — so **`(3,20)` and `(4,30)` cannot be produced through the public API**.

Restricted to the three reachable rows: χ² `0.057 / 0.057 / 0.058` against F `0.054 / 0.056 / 0.055` — a separation of **≤ 0.003**. The TL;DR's `5.7–6.7%` upper bound of `0.067` *is* the `T=20` row. The number carrying the change is one users cannot reach.

This is the same shape as #813's first version (`6.8% at T = 8`, also below the floor). Worth adopting as a habit: run the measurement grid through `_require_slice_floor` before quoting it.

### At reachable sizes the correction is inert

Measured independently through the public path (`build_disjoint_period_panel` + `ic()`, real HAC variance estimation):

| K | T | χ² (main, measured) | F (this PR) | mean ν |
|---|---|---|---|---|
| 3 | 50 | 0.080 | 0.074 | 138 |
| 3 | 90 | 0.058 | 0.058 | 257 |
| 3 | 150 | 0.060 | 0.060 | 434 |
| 5 | 50 | 0.078 | 0.076 | 227 |
| 5 | 90 | 0.078 | 0.074 | 425 |
| 5 | 150 | 0.080 | 0.080 | 720 |

At most 0.006, and exactly 0.000 in three of six cells — because **ν lands between 138 and 720**, where `F_{K−1, ν}` and `χ²_{K−1}/(K−1)` are the same distribution for practical purposes. (Reconstructing the old reference as `chi2.sf(stat, K−1)` on the same stats reproduces `origin/main` exactly, so the harness is sound.)

My absolute level is higher than the PR's because the fixtures differ — idealised iid slices there, the full public path with real HAC variance estimation here. Both are valid; the second is what users see.

### The larger issue, not introduced here

Under either reference the joint path sits at **5.8–8.0%** and **does not converge in T** — K=5 at T=150 is still 0.080. With ν in the hundreds, no choice of denominator df can fix that. It means the over-rejection is not the reference distribution but the Wald statistic itself being too large, i.e. the HAC variances from `_analytic_slice_moments` are understated.

Pre-existing, but the PR as written reads as though it addresses it.

### Suggested resolution

**Keep the change, rewrite the justification, and open an issue for the variance understatement.** `F_{K−1, ν}` is still the more defensible reference on principle (finite sample, estimated `V`), it makes the joint path consistent with the pairwise one, and it discloses ν — exactly the intra-file inconsistency #803 complained about. That is an honest rationale. "Size improvement" is not, at reachable sizes.

Concretely: restrict the table to reachable cells, retitle the rationale as consistency + disclosure, and state plainly that the residual 5.8–8.0% and the K=5 non-convergence are unresolved and point at `_analytic_slice_moments`.

The author of `fd754bb` has agreed with this disposition.

Happy to push the reworded rationale, or verify it once you have — your call; I do not touch other authors' branches unasked.

### On direction, since it was asked here

I'd take **(a)**, with **(c) reframed**. The disclosed iid 6.7% in #815 is the smaller miscalibration. Measured, the OLS SE it replaced rejects at **33%** under AR(0.6) residuals, and HAC — while far better — still sits at 11.5–18.5% and converges slowly (0.185 → 0.115 from n=60 to n=240). That is a Bartlett small-sample property, so every HAC path in the library inherits it, not just spanning. So (c) is not a spanning fix; it is a library-wide calibration project.

Proposal: fold them into one step. Run the (a) sweep, but include "empirical size of every HAC path under autocorrelated input" in its scope. That table then decides whether (c) is worth the Kiefer-Vogelsang / Andrews-Monahan level of effort, on evidence rather than blind. `n/(n−k)` is already ruled out — I measured it at 0.071 → 0.066, inside the noise. #575 is worth a ten-minute look first, to see whether it holds numeric items that belong in the same sweep.

This needs the user's go-ahead before either of us starts; I have not begun.

awwesomeman:
## Direction settled — take (a). Go ahead and start.

The user has signed off, so this is a green light, not a proposal.

**Tracker: #821.** The variance understatement I found while reviewing this PR is tracked separately as **#820** — that one is a live calibration defect, not a characterisation task.

### Why (a) over (c), on mainstream grounds

(c) as framed targets the wrong number. The disclosed iid 6.5% in #815 is the smaller miscalibration; measured, the OLS SE it replaced rejects at 33% under AR(0.6) residuals while HAC — far better — still sits at 11.5–18.5% and converges slowly (0.185 → 0.115 from n=60 to n=240). That is a Bartlett small-sample property, so it is not a `spanning` fix; every HAC path in the library inherits it.

More to the point, replacing the kernel is not what the field does:

- Reporting Newey-West t-stats and disclosing their short-sample behaviour is the dominant convention in empirical asset pricing. NW over-rejection in short samples is well documented, not a new finding.
- The standard responses are to **raise the hurdle** (Harvey-Liu-Zhu 2016's t > 3.0) or to **switch to a bootstrap**, not to swap the kernel.
- At library level: `statsmodels` offers an optional `n/(n−k)` correction and does not do fixed-b by default; `arch` goes the bootstrap route. Fixed-b is a research-literature tool, not a library default.

factrix already ships bootstrap paths (`STATIONARY_BOOTSTRAP`, `_block_bootstrap_diff_p`, the bootstrap omnibus/pairwise). So the most valuable output of the sweep is probably **routing guidance** — "below n = X, or above persistence Y, prefer the bootstrap path" — which costs almost nothing to implement and matches what practitioners actually do. That is a better deliverable than a kernel rewrite, and the sweep is what tells us where the thresholds are.

If the table does justify a kernel-level fix afterwards, prefer **Andrews-Monahan (1992) prewhitening** over Kiefer-Vogelsang fixed-b: prewhitening changes only the SE and is a drop-in inside the kernel, leaving downstream p-values, BHY and the gates untouched, whereas fixed-b changes *critical values* and collides with the whole p-value pipeline. `n/(n−k)` is already ruled out — I measured 0.071 → 0.066, inside the noise.

### One discipline point

Run the measurement grid through each metric's floor before quoting any cell. #813 and #819 each drew a conclusion from a configuration the public API cannot produce. That is the single cheapest guard against repeating this class of finding.

### On this PR

Unchanged from my review above: keep the change, rewrite the rationale as consistency + disclosure, restrict the table to reachable cells, and state that the residual 5.8–8.0% is unresolved and tracked in #820. Once that lands I will verify and merge.

PR #900, opened 2026-08-28
feat(preprocess): warn on an uneven evaluation grid

awwesomeman:
> **TL;DR** — `compute_forward_return(dates=)` now fires a new
> `uneven_evaluation_grid` warning when the kept evaluation rows are not a
> constant number of periods apart, naming the paths that assume a constant
> spacing and the caller-side remedy. Disclosure only: no metric behaviour,
> no new parameters, no resampling.

## Summary

- New `WarningCode.UNEVEN_EVALUATION_GRID`, filed with the preprocess /
  data-shape codes next to `ragged_period_grid`. Its description names the
  paths that read a constant spacing into the grid they are handed — the
  regression HAC tests (`predictive_beta`, `spanning_alpha`, the
  K-restriction Wald / slice period joint tests), the ADF / autocorrelation
  persistence screens, event-study estimation windows and offsets, and
  adjacent-period metrics (turnover / rank autocorrelation, rolling
  windows) — and states that the remedy is the caller's, since factrix does
  not resample.
- `compute_forward_return` emits it from a `_warn_if_uneven` helper next to
  `_warn_if_ragged`, at the site where the overlap is derived from the kept
  index, and only when `dates=` was given. The message carries the one
  secondary note: on an uneven grid `overlap_periods` is the *maximum*
  overlap, so `NonOverlapping` strides at that maximum and discards more of
  the series than on a constant-stride grid, while `NeweyWest`'s
  `1.3*sqrt(T)` base bandwidth is insensitive to it — sample efficiency,
  not a way to recalibrate the paths above, which `inference=` cannot
  switch.
- Docs: the existing uneven-grid disclosure in
  `docs/reference/statistical-methods.md` gains the bulleted list of
  constant-spacing assumptions, the remedy and the new code;
  `docs/guides/preparing-data.md` points at the warning from the bullet
  that introduces `dates=`.

## Test plan

- [x] `tests/test_evaluation_grid.py::TestUnevenEvaluationGridWarning` —
      a `(20, 20, 40)`-spaced grid warns with the code in the message; a
      constant-stride grid and the full grid do not; the code has a
      description.
- [x] `uv run pytest -q -x` — 3070 passed, 42 skipped.
- [x] `uv run ruff check` / `ruff format --check` / `mypy factrix` clean.
- [x] `uv run mkdocs build --strict` clean; the regenerated
      `docs/reference/_generated_warning_codes.md` row is committed.

Refs #862 — this is the disclosure half of that tracker's parked
"inference calibration on an uneven grid" item. The Monte Carlo re-measurement
of the regression HAC paths stays parked; nothing here changes a number.


awwesomeman:
> **TL;DR** — Mechanism is exactly parallel to `ragged_period_grid` (same site, same shape, no metric wiring), tests cover uneven / constant-stride / full-grid, docs name the four constant-spacing assumptions and the caller-side remedy. Two trims before merge.

1. **Runtime warning text is too long — cut it to the `_WARNING_DESCRIPTIONS` entry.** `_warn_if_ragged` is three sentences; this one is a paragraph and carries a simulation number (`5.5%`) that only the stats page otherwise pins. Drop everything from "Separately, on an uneven grid overlap_periods is the maximum overlap…" to the end of the message. The `NonOverlapping`-vs-`NeweyWest` sample-efficiency note already lives in `statistical-methods.md`, which is where a number belongs; the runtime message should be: what was detected (keep the `{N} distinct spacings ({min} to {max} periods)` detail — that part is good), which paths assume a constant spacing, and the remedy. Also use "the caller's side" rather than "your side" to match the description entry.
2. **Drop `test_uneven_grid_code_is_documented`.** `WarningCode.description` is a dict lookup over the whole enum; the existing description-coverage test (and the generated `_generated_warning_codes.md` hook, which wrote 40 rows) already fail if an entry is missing. The test asserts nothing the suite does not already guarantee.

Fine as-is: the emit condition (any ≥2 distinct spacings, including a single odd date) is right — one odd date is enough to break every assumption listed, and the pre-existing max-overlap test in the same file shows exactly that; the `catch_warnings` wrap on that test matches the file's style.

PR #1034, opened 2026-09-02
fix(metrics): report the HAR bandwidth the kernel ran at

awwesomeman:
## Summary

`predictive_beta` resolves its Bartlett bandwidth against the full finite-pair
count `n`, then fits the Amihud-Hurvich augmented design on `m = n - h` rows.
A Bartlett sum cannot use a lag it has no observation pair for, so
`_ols_nw_multivariate` clipped to `m - 1` internally while
`metadata["har_lags"]` kept reporting the wider request.

- `AmihudHurvichFit` carries `lags_used`, the bandwidth the covariance was
  actually computed at
- `predictive_beta` reports that instead of the resolved request
- `_har_dof` reads the same value
- the ill-conditioned-bandwidth screen keeps its calibrated condition on the
  resolved bandwidth and now names the applied one when the two differ, so its
  text cannot contradict `metadata["har_lags"]`

Scope note (from review): this closes the `har_lags` contradiction only. The
same message still names `n_periods` while printing `n_periods_finite`, which
diverges from `metadata["n_periods"]` on this metric — filed as #1035 rather
than folded in, because choosing which count the message should name is a
separate wording decision from reporting the applied bandwidth.

## Reachability

Through the public API, no explicit bandwidth needed:

| n | h | reported `har_lags` | `n_periods` | bandwidth the kernel used |
|---|---|---|---|---|
| 60 | 42 | 20 | 18 | 17 |
| 60 | 50 | 20 | 10 | 9 |
| 75 | 63 | 25 | 12 | 11 |
| 90 | 63 | 30 | 27 | 26 |

Each row is a real test, not a short circuit. `n_periods=27` with
`har_lags=30` is a bandwidth wider than the sample it ran on.

## Quant impact

None. The standard error was already computed at the clipped bandwidth, so no
SE changes. `_har_dof` now reads `lags_used`, but the `n/h - 1` cap floors the
effective df at 1.0 in every currently reachable cell, so no p-value moves
either — verified across the four cells above. This corrects what is
**reported**, and removes a latent path where a future bandwidth or floor
change would silently move the df.

The sibling paths already behave this way, which is what makes this an
outlier rather than a convention: `_ols_scalar_wald_hac` resolves against
`len(y)` — the design it fits — and `_driscoll_kraay_cov` clips before it
returns `lags_used`.

## How it was found

Cross-validating the statistical kernels against `statsmodels` 0.15 and
`scipy` 1.17. The estimators themselves matched to machine precision:
Bartlett long-run variance and `K`-vector long-run covariance against
`OLS(cov_type="HAC")`, `_ols_nw_multivariate` (coefficients, HAC covariance,
residuals), `_driscoll_kraay_cov` against `cov_nw_groupsum`,
`_wald_two_way_cluster` against a CGM reference built with the corrections
and `F` reference the docstring documents, `_adf` against `adfuller`,
`_ljung_box` against `acorr_ljungbox`, `_binomial_two_sided_p` against
`binomtest`, `holm_adjusted_p` / `bhy_adjusted_p` against `multipletests`,
plus `_lag1_autocorr`, `_mann_kendall_hamed_rao` (tau against
`scipy.stats.kendalltau`), `_batch_means_se` and `_empirical_p` against
closed forms. This bandwidth report was the only discrepancy.

## Validation

- `pytest -p no:faulthandler -q` — 3795 passed, 45 skipped (main: 3789; +6 new)
- `ruff check factrix tests`
- `ruff format --check factrix tests`
- `mypy factrix`
- `mkdocs build --strict`

The six new tests fail on `main`: three on `assert har_lags <= n_periods - 1`
(`assert 30 <= 26`, `assert 25 <= 11`, `assert 20 <= 17`) and three on the
absent `lags_used` field.

### DIFF
--- a/docs/api/metrics/predictive_beta.md
+++ b/docs/api/metrics/predictive_beta.md
@@ -42,9 +42,11 @@ title: factrix.metrics.predictive_beta
     The raw OLS reference retained in metadata stays on the narrower
     Newey-West bandwidth and is labelled as uncorrected.
     Metadata reports the headline branch directly: `hac_applied=False` and
-    `har_lags=None` at `h = 1`; both switch to the resolved HAR path at
-    `h > 1`. The `method` string names the covariance that produced the
-    headline test.
+    `har_lags=None` at `h = 1`; both switch to the HAR path at `h > 1`. The
+    `method` string names the covariance that produced the headline test, and
+    `har_lags` is the bandwidth the kernel ran at — the augmented design is
+    `n_periods` rows, so a long horizon can leave it narrower than the
+    bandwidth resolved from the full series.
 
 -   __Persistent predictor diagnostic__
 
--- a/docs/reference/stat-keys-by-metric.md
+++ b/docs/reference/stat-keys-by-metric.md
@@ -910,8 +910,11 @@ which reports `R²` for the OLS regression.
   `residual_lag1_autocorr` (lag-1 autocorrelation of the reported model's
   residuals, strided at `overlap_periods`), `newey_west_lags`,
   `overlap_periods`, `hac_applied` (whether the headline corrected test used
-  HAC), `har_lags` (the headline HAR bandwidth at `overlap_periods > 1`;
-  `None` at `h = 1`, where the corrected test is homoskedastic;
+  HAC), `har_lags` (the headline HAR bandwidth the kernel **ran at** at
+  `overlap_periods > 1` — the augmented design is `n_periods` rows and a
+  Bartlett sum cannot use a lag it has no pair for, so on a long horizon this
+  is narrower than the bandwidth resolved from the full series; `None` at
+  `h = 1`, where the corrected test is homoskedastic;
   `newey_west_lags` stays the narrow covariance bandwidth resolved for the
   uncorrected OLS reference), `alpha` (the intercept the corrected slope implies
   on the `n_obs` rows), `r_squared_ols_uncorrected`, `factor_std`,
--- a/factrix/_stats/ols.py
+++ b/factrix/_stats/ols.py
@@ -300,8 +300,17 @@ class AmihudHurvichFit(NamedTuple):
     the augmented regression's intercept: that one absorbs the innovation
     proxy and is not the intercept implied by ``beta``.
 
-    A not-computable fit fills every float with NaN, ``n_used`` with ``0``
-    and ``resid`` with an empty array.
+    ``lags_used`` is the Bartlett bandwidth the covariance was **actually**
+    computed at, which is not always the ``lags`` the caller asked for: the
+    augmented design runs on ``n_used`` rows and a kernel cannot use a lag it
+    has no observation pair for. Callers report this rather than their own
+    request, the way :func:`_ols_scalar_wald_hac` and
+    :func:`~factrix._stats.hac._driscoll_kraay_cov` already report the
+    bandwidth they ran at. It is ``0`` at ``overlap_periods = 1``, where the
+    covariance is homoskedastic and no kernel runs.
+
+    A not-computable fit fills every float with NaN, ``n_used`` and
+    ``lags_used`` with ``0`` and ``resid`` with an empty array.
     """
 
     beta: float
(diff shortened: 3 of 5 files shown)


awwesomeman:
## Independent review — approved

Reviewed in a detached worktree at `76cf3548`, re-deriving the claims rather than
reading them off the PR text.

### 1. "No p-value moves" — confirmed, and by a stronger mechanism than claimed

Swept every reachable `(n, h, explicit-bandwidth)` cell (`n` to 1500, `h` in
`2..n-1`, `m >= 4`, user bandwidths `None` plus 1/2/3/5/12/30/100/`n/2`/`n-1`/`n`/`10n`)
comparing `_har_dof(m, lags, h)` against `_har_dof(m, lags_used, h)`:

```
cells where the kernel clipped (lags_used < lags): 4454851
cells where _har_dof MOVES:                        0
min resolved-lags/m ratio = 1.0000    max(m/h) = 0.6667
```

The PR credits the `n/h - 1` cap. That holds — `max(m/h) = 0.667` over clip
cells, so the cap term `m/h - 1` is *negative* there. But there is a second,
independent reason worth recording, because it is structural rather than
calibrated:

- Clipping happens iff `lags > m - 1`, i.e. `lags >= m`. So the LLSW term
  `1.5·m/lags - 1 <= 0.5 < 1` — the **old** df was already floored to 1.0 by
  itself, with no help from the cap.
- For the **new** df, `1.5·m/(m-1) - 1 > 1` requires `m < 4`, and `m >= 4` is
  forced by the 3-column design's `n >= k+1` guard.

So both values are pinned at 1.0 whenever the switch could matter, and that does
*not* depend on `_MAX_BANDWIDTH_FRACTION`, the `3(h-1)` floor, or any explicit
bandwidth surviving `_resolve_har_lags`. The "removes a latent path where a
future bandwidth change would silently move the df" framing is right, and the
invariant is more robust than the PR states.

`lags_used = max(0, min(lags, m - 1))` binds exactly to the kernel's own
`L = max(0, min(lags, n - 1))` in `_ols_nw_multivariate_from_finite`, so the SE
is genuinely unchanged. This is a reporting-only change. Confirmed.

### 2. The new tests are not empty assertions

Reverted `factrix/_stats/ols.py` and `factrix/metrics/predictive_beta.py` to
`23015f54` while keeping the new test file. All 6 fail, on the stated values:

```
E   assert 20 <= (18 - 1)
E   assert 25 <= (12 - 1)
E   assert 30 <= (27 - 1)
E   AttributeError: 'AmihudHurvichFit' object has no attribute 'lags_used'   (x2)
E   assert 'the kernel ran at L=30' in '...the resolved Bartlett bandwidth L=30...'
6 failed
```

### 3. Warning scope and clause truth — agree, and verified

Keeping the firing condition on `(n, resolved_har_lags)` is the right scope
line: the screen was calibrated on the finite-pair count, and moving it to the
truncated design is a recalibration, not a reporting fix.

The appended clause is always true when it fires. Over 859 cells (`n` 40..200,
`h` 2..n-4) the screen fired 797 times with **0** false or inconsistent clauses.
The clause only appears when `resolved > m - 1`, which forces
`lags_used = m - 1 = n_used - 1` exactly as the text asserts.

### Verification

`3795 passed, 45 skipped` (main is 3789, +6). `ruff check`, `ruff format --check`,
`mypy factrix` all clean. CI: 17 pass / 2 skipping, `mergeStateStatus CLEAN`.
`AmihudHurvichFit` is constructed only inside `_amihud_hurvich_beta`, so
appending a field breaks no positional unpacking.

### Two follow-ups — filed separately, not blocking

Neither is caused by this PR; both are adjacent to the contract it tightens, so
per repo convention they get their own issues rather than scope creep here.

1. **The same warning still contradicts metadata on a different token.** The
   message reads `exceeds n_periods / 5 on 90 periods`, but `metadata["n_periods"]`
   is `27` — the message's `n_periods` is `metadata["n_periods_finite"]`. Since
   this PR's stated goal is that the text cannot contradict the metadata, the
   remaining half is worth closing.

2. **The short-circuit branch reports a bandwidth no kernel ran at.** Leaving
   `har_lags=resolved_har_lags` at line 336 was defensible before; now that the
   docstring says `har_lags` *is* the bandwidth the kernel ran at, that branch
   contradicts its own documentation. It is reachable — 560 cells in a small
   sweep (e.g. `n=20, h=19` reports `har_lags=7` on a 1-row design), and
   `n_periods` is `None` there, so a consumer cannot derive the bound either.

PR #1038, opened 2026-09-02
fix(metrics): name the metadata key the bandwidth screen reads

### LINKED ISSUE #1035: predictive_beta's ill-conditioned-bandwidth warning names n_periods but reports n_periods_finite
Split out of the review of #1034, which closed the other half of this contradiction.

`predictive_beta` emits:

    the resolved Bartlett bandwidth L=30 exceeds n_periods / 5 on 90 periods,
    so the long-run variance rests on too few lag products to be stable.

with metadata:

    {'har_lags': 26, 'n_periods': 27, 'n_periods_finite': 90, 'newey_west_lags': 62}

The message's `n_periods` token is `metadata["n_periods_finite"]` (90), not
`metadata["n_periods"]` (27). A reader who takes the warning at its word and
looks up `n_periods` gets 27 and cannot reproduce the `90 / 5` comparison.

The screen itself is correct — `_hac_bandwidth_ill_conditioned(n, resolved_har_lags)`
is calibrated on the finite-pair count, and #1034 deliberately left that firing
condition alone. This is purely the message naming the wrong metadata key.

`predictive_beta` is the only metric where the two diverge: its
`metadata["n_periods"]` is the truncated augmented-design row count `n_used`,
whereas `fm_beta` and the shared `factrix/metrics/_helpers.py` phrasing pass
the same value they report.

### In-boundary assessment

In boundary for a message-wording fix: say `n_periods_finite` (or state the
count without naming a key). Out of boundary for anything that moves the
screen's firing condition onto the truncated design — that is a recalibration,
not a reporting fix.

Ref: `factrix/metrics/predictive_beta.py` (the `HAC_BANDWIDTH_ILL_CONDITIONED`
emit), `factrix/metrics/_helpers.py`.

### LINKED ISSUE #1036: predictive_beta short circuit reports a har_lags no kernel ran at
Split out of the review of #1034.

#1034 tightened the `har_lags` contract to "the bandwidth the kernel ran at"
in both the docstring and `docs/reference/stat-keys-by-metric.md`. The
`no_amihud_hurvich_fit` short circuit still reports
`har_lags=resolved_har_lags`, and on that branch no kernel runs at all, so the
value contradicts the contract it is now documented under.

Not a regression — main behaved the same way before #1034 — but the tightened
wording is what turns it into an inconsistency.

### Reachable

560 cells in a small sweep (`n` 20..160, `m = n - h <= 6`). Examples:

| n | h | m (design rows) | reported `har_lags` | `n_periods` |
|---|---|---|---|---|
| 20 | 19 | 1 | 7 | `None` |
| 20 | 16 | 4 | 7 | `None` |
| 22 | 18 | 4 | 8 | `None` |

`n=20, h=19` reports a bandwidth of 7 for a one-row design. `n_periods` is
`None` on this branch, so a consumer cannot derive the `n_used - 1` bound for
itself either.

### Options

1. Report `None` — but `docs/reference/stat-keys-by-metric.md` currently
   documents `None` as meaning `h = 1`, so that overloads the sentinel.
2. Keep the resolved value and document the short circuit as an explicit
   exception ("the bandwidth that would have been used").
3. Carry `n_periods` through the short circuit so the value is at least
   interpretable.

Option 2 is the smallest change and keeps the `None`-means-`h=1` reading
intact; option 3 is a separate metadata-completeness question.

### In-boundary assessment

In boundary: the one metadata value and the doc sentence that describes it.
Out of boundary: the broader question of which keys every `predictive_beta`
short circuit should populate.

Ref: `factrix/metrics/predictive_beta.py`, the `no_amihud_hurvich_fit`
`_short_circuit_output` call.

awwesomeman:
Two contracts left open by the review of #1034, fixed together because they are
adjacent and both touch the same `har_lags` documentation sentence.

## #1035 — the message named a key that gives a different number

`predictive_beta` emitted:

    the resolved Bartlett bandwidth L=30 exceeds n_periods / 5 on 90 periods, ...

with `metadata["n_periods"] == 27` and `metadata["n_periods_finite"] == 90`. A
reader who looked up the key the message named could not reproduce the `90 / 5`
comparison. This metric is the only place the two counts diverge: `n_periods`
is the truncated augmented design's row count, while the screen is calibrated
on the finite pairs.

The message now names `n_periods_finite`. **The firing condition is
unchanged** — `_hac_bandwidth_ill_conditioned(n, resolved_har_lags)` still
reads the finite-pair count, because moving the screen onto the truncated
design is a recalibration rather than a reporting fix.

## #1036 — the short circuit runs no kernel

#1034 tightened `har_lags` to "the bandwidth the kernel ran at". The
`no_amihud_hurvich_fit` short circuit cannot satisfy that, since no kernel runs
there at all. Per the issue's recommended option, the value is kept as the
bandwidth resolved for the attempt and the exception is documented, rather than
reusing `None` — which already means `h = 1` and would be overloaded. The
behaviour is now pinned by test so it cannot drift silently in either
direction.

No new metadata key: `metadata["reason"]` is already the machine-readable
signal for this branch, and the issue scoped "which keys every short circuit
should populate" as out of boundary.

## Test that had to change

`test_bandwidth_warning_text_uses_the_policy_constant` asserted that *every*
bandwidth message carries the literal `n_periods / 5`. Its real intent is that
`_MIN_PERIODS_PER_LAG` is interpolated rather than hand-written — the key name
was an incidental assumption, and #1035 is precisely the finding that the
assumption is wrong for this metric. The shared `_INFERENCE_CODE_MESSAGE`
template and `_emit_scalar_har_warnings` keep the full token; `predictive_beta`
is asserted to carry the interpolated constant *and* to name
`n_periods_finite`, so the check did not get weaker.

Flagging it explicitly because loosening an existing consistency test is the
kind of change that deserves a second opinion.

## Also

The doubled "ran at at" that #1034 introduced in `stat-keys-by-metric.md`,
folded in here per the review rather than spending its own CI cycle.

## Validation

- `pytest -p no:faulthandler -q` — 3805 passed, 45 skipped (main: 3795; +10 new)
- `ruff check` / `ruff format --check` / `mypy factrix`
- `mkdocs build --strict`, no generated-file drift

Five of the ten new tests fail on `main`: three on the message naming
`n_periods` where the printed count is `n_periods_finite`, one on the bare
`n_periods /` token, and one on the absent short-circuit doc sentence. The
other five pin behaviour that is already correct and must stay that way.

Closes #1035
Closes #1036

### DIFF
--- a/docs/reference/stat-keys-by-metric.md
+++ b/docs/reference/stat-keys-by-metric.md
@@ -910,11 +910,14 @@ which reports `R²` for the OLS regression.
   `residual_lag1_autocorr` (lag-1 autocorrelation of the reported model's
   residuals, strided at `overlap_periods`), `newey_west_lags`,
   `overlap_periods`, `hac_applied` (whether the headline corrected test used
-  HAC), `har_lags` (the headline HAR bandwidth the kernel **ran at** at
-  `overlap_periods > 1` — the augmented design is `n_periods` rows and a
+  HAC), `har_lags` (at `overlap_periods > 1`, the headline HAR bandwidth the
+  kernel **ran at** — the augmented design is `n_periods` rows and a
   Bartlett sum cannot use a lag it has no pair for, so on a long horizon this
   is narrower than the bandwidth resolved from the full series; `None` at
-  `h = 1`, where the corrected test is homoskedastic;
+  `h = 1`, where the corrected test is homoskedastic. One exception: on the
+  `no_amihud_hurvich_fit` short circuit no kernel runs at all, and the value
+  is the bandwidth resolved for the attempt. `None` is not reused there
+  because it already means `h = 1`;
   `newey_west_lags` stays the narrow covariance bandwidth resolved for the
   uncorrected OLS reference), `alpha` (the intercept the corrected slope implies
   on the `n_obs` rows), `r_squared_ols_uncorrected`, `factor_std`,
--- a/factrix/metrics/predictive_beta.py
+++ b/factrix/metrics/predictive_beta.py
@@ -151,7 +151,14 @@ def predictive_beta(
     the kernel ran at: the augmented design is ``n_periods`` rows and a
     Bartlett sum cannot use a lag it has no observation pair for, so a long
     horizon leaves it narrower than the bandwidth resolved from the full
-    series.
+    series. The one exception is the ``no_amihud_hurvich_fit`` short circuit,
+    where no kernel runs and the value is the bandwidth resolved for the
+    attempt; ``None`` is not reused there because it already means ``h = 1``.
+
+    The ``HAC_BANDWIDTH_ILL_CONDITIONED`` message names
+    ``n_periods_finite``, not ``n_periods``. This metric is the one place the
+    two diverge — the screen is calibrated on the finite pairs, while
+    ``n_periods`` counts the truncated augmented design.
 
     **The correction costs power** where OLS's apparent power was partly
     its own bias: at $T=60,\ \phi=0.95,\ \rho=-0.9$ the corrected test
@@ -405,10 +412,15 @@ def predictive_beta(
         )
         _emit_warning(
             WarningCode.HAC_BANDWIDTH_ILL_CONDITIONED,
+            # Name ``n_periods_finite``, not ``n_periods``. This metric is the
+            # one place the two diverge: ``n_periods`` is the truncated
+            # augmented design's row count, while the screen reads the finite
+            # pairs. A bare "n_periods" sends a reader to a number that cannot
+            # reproduce the comparison the message states.
             f"the resolved Bartlett bandwidth L={resolved_har_lags} exceeds "
-            f"n_periods / {_MIN_PERIODS_PER_LAG} on {n} periods, so the "
-            f"long-run variance rests on too few lag products to be "
-            f"stable.{narrowed}",
+            f"n_periods_finite / {_MIN_PERIODS_PER_LAG} on the {n} finite "
+            f"pairs, so the long-run variance rests on too few lag products "
+            f"to be stable.{narrowed}",
             label="predictive_beta",
             expected_warnings=expected_warnings,
             warning_codes=warning_codes,
--- a/tests/stats/test_predictive_beta_metadata_contract.py
+++ b/tests/stats/test_predictive_beta_metadata_contract.py
@@ -16,6 +16,7 @@ Two contracts, both split out of the review of #1034:
 from __future__ import annotations
 
 import math
+import re
 import warnings
 from pathlib import Path
 
@@ -99,6 +100,15 @@ class TestShortCircuitBandwidthIsTheAttemptedOne:
         assert "the bandwidth resolved for the attempt" in docs
 
     def test_docs_do_not_repeat_at_at(self) -> None:
-        """Typo introduced by #1034's wording; folded in here per review."""
+        """Typo introduced by #1034's wording; folded in here per review.
+
+        The literal ``"ran at at"`` never appeared: #1034 shipped
+        ``the kernel **ran at** at``, with the emphasis markers between the
+        two words. A bare-substring guard passed while the typo was live, so
+        match across the markup instead.
+        """
         docs = STAT_KEYS_DOCS.read_text(encoding="utf-8")
-        assert "ran at at" not in docs
+        doubled = re.compile(r"\bat(?:\*+|_+|`+|\s)*\s+at\b")
+        assert not doubled.search(docs), (
+            f"doubled 'at' in {STAT_KEYS_DOCS}: {doubled.search(docs).group()!r}"
+        )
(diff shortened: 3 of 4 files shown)


awwesomeman:
## Independent review — two one-line changes requested, then good to merge

Reviewed in a detached worktree at `5dd8025c`.

### The loosened consistency test: your reading is right, but it lost adjacency

You asked whether this preserves the invariant or erodes it. **The invariant you
named is preserved.** Both realistic regressions are still caught:

- constant hand-written instead of interpolated → if `_MIN_PERIODS_PER_LAG`
  moves to 6, `policy_token` becomes `"/ 6"` and a message still saying `/ 5`
  fails;
- key name silently reverting → `"n_periods_finite /"` is asserted literally, so
  a message going back to `n_periods /` fails.

So this is not the erosion you were worried about. But splitting one token into
two independent `in` checks does drop something the original had: the constant
is no longer verified to be *the divisor of the named key*. Demonstrated:

```
message: "... exceeds n_periods_finite / 9 on the 90 finite pairs, so the
          long-run variance rests on too few lag products / 5 to be stable."

  PR's two split assertions accept it?  True
  single combined token accepts it?     False
```

Contrived, but the fix is strictly stronger and costs one line — collapse the
two back into the same shape as the original, with the corrected key:

```python
assert all(
    f"n_periods_finite / {_MIN_PERIODS_PER_LAG}" in message
    for message in predictive_messages
)
```

That is exactly the original invariant with the key name #1035 corrected, rather
than a weaker pair. Please make this change.

### `test_docs_do_not_repeat_at_at` cannot detect the typo it is named for

This is the one pinning test I'm not happy with, and it fails your own standard
rather than mine. It asserts `"ran at at" not in docs`. The typo #1034 actually
introduced was:

```
the headline HAR bandwidth the kernel **ran at** at `overlap_periods > 1`
```

The literal `ran at at` never appears — the bold markers sit between the two
`at`s. Verified against `f418d4b2`:

```
occurrences of "ran at at"   in main docs: 0
occurrences of "ran at** at" in main docs: 1
```

So the guard passes on main, would have passed *while the typo was live*, and
will not catch a reintroduction in the form it actually took. The other four
pinning tests are fine — `test_short_circuit_keeps_the_resolved_bandwidth` and
`test_h1_still_reports_no_bandwidth` genuinely fail if someone swaps the
sentinel, which is the point of pinning #1036's decision. This one is the
exception. Please assert against the form the typo really takes, e.g.
`"ran at** at" not in docs`, or a regex tolerant of the markup.

### On the alternative you offered — don't take it

Renaming `metadata["n_periods"]` on this metric would remove the special case,
and pre-1.0 it is allowed. I still think you were right not to reach for it: it
trades a one-line message fix for a breaking change to a key that reads
consistently across every other metric, and #1035 was scoped to the message
naming the wrong key, not to the metadata vocabulary. If the divergence is worth
revisiting it should be a deliberate metadata-naming decision across all the
metrics, not a side effect of a warning-text fix.

### Secondary questions

1. **Agreed the `har_lags_applied` flag would have been redundant.**
   `metadata["reason"] == "no_amihud_hurvich_fit"` already identifies the branch
   exactly, and a second key encoding the same fact is one more thing to keep in
   sync. Dropping the draft was right.

2. **`hac_applied` is a real instance, and it wants its own issue.** Confirmed
   on the short circuit at `n=20, h=19`:

   ```
   reason      = 'no_amihud_hurvich_fit'
   har_lags    = 7
   hac_applied = True
   n_periods   = None
   method      = None
   ```

   `hac_applied` is documented as whether the headline corrected test used HAC,
   and on that branch there is no headline corrected test — `method` is `None`
   in the same payload, which is the contradiction in plain sight. Same shape as
   `har_lags`, outside both issues, so a quiet fix here would be exactly the
   scope creep the convention exists to prevent.

### Verification

`3805 passed, 45 skipped` (main 3795, +10). `ruff check`, `ruff format --check`,
`mypy factrix` clean. CI 17 pass / 2 skipping, `mergeStateStatus CLEAN`.

Reverting `predictive_beta.py` and `stat-keys-by-metric.md` to `f418d4b2` while
keeping the new tests: 5 of the new tests fail, plus the modified consistency
test — your count of 5 new is right, the 6th is the existing test you edited,
which is expected. And yes, pinning tests that pass on main are fine as guards;
that is what they are for, with the `at at` exception above.

awwesomeman:
## Re-verified at `5712a9be` — both fixes good, merging

**Fix 1 (combined token).** Took the tighter form; the counterexample that passed the split assertions now fails it.

**Fix 2 (markup-tolerant regex).** Verified independently rather than on the diff:

```
regex flags f418d4b2 (the shipped typo)?  True   'at** at'
regex clean on the corrected file?        True
old substring guard would have caught it? False
false-positive sweep over all docs/**/*.md: 1 match — the typo itself
```

So the new guard demonstrably catches the defect the old one could not, and fires nowhere else in the docs tree.

**Validation at `5712a9be`:** `3805 passed, 45 skipped`; `ruff check`, `ruff format --check`, `mypy factrix` clean; CI 17 pass / 2 skipping, `mergeStateStatus CLEAN`.

`hac_applied` on the short circuit is tracked separately as #1039, correctly kept out of this PR.

PR #1040, opened 2026-09-02
fix(metrics): state the short-circuit metadata contract, judge keys per side

### LINKED ISSUE #1039: fix(metrics): predictive_beta short circuit reports hac_applied=True with method=None
Raised during the review of #1038, outside the boundary of both #1035 and
#1036, so filed rather than fixed quietly there.

`hac_applied` is documented in `docs/reference/stat-keys-by-metric.md` as
"whether the headline corrected test used HAC". On the
`no_amihud_hurvich_fit` short circuit there is no headline corrected test at
all, and the key still reports `True`.

Reproduced:

```
n=20 h=19: reason='no_amihud_hurvich_fit' hac_applied=True method=None har_lags=7 n_periods=None
n=20 h=16: reason='no_amihud_hurvich_fit' hac_applied=True method=None har_lags=7 n_periods=None
n=22 h=18: reason='no_amihud_hurvich_fit' hac_applied=True method=None har_lags=8 n_periods=None
```

`method=None` in the same payload is the contradiction in plain sight: the
metadata simultaneously says no covariance method produced a result and that
the headline test used HAC.

This is the same shape as #1036 — a key whose contract describes a completed
test, reported on a branch where none ran — one key over. #1023 added
`hac_applied`; #1034 tightened `har_lags`; #1038 documented the `har_lags`
exception. `hac_applied` was not revisited in any of them.

### Why it is not simply "set it to False"

`hac_applied` is currently `overlap_periods > 1`, a statement about the
*branch that would have run*, and the success path relies on that reading:
`har_lags` is `None` at `h = 1` precisely because `hac_applied` is `False`
there. Flipping it to `False` on the short circuit would make `hac_applied`
mean two different things depending on `reason`, and would make the short
circuit indistinguishable from `h = 1` for a consumer reading the pair.

The options mirror #1036's:

1. Leave the value and document the exception, as #1038 did for `har_lags`.
   Cheapest and consistent with the precedent just set, but it is a second
   documented exception on the same branch, which starts to argue that the
   branch — not the individual keys — is what needs a contract.
2. State the branch contract once: on `no_amihud_hurvich_fit`, every
   inference-describing key (`method`, `hac_applied`, `har_lags`,
   `n_periods`) describes the attempt, not a result. One sentence replacing
   per-key exceptions.
3. Populate `n_periods` on the branch so the values are at least
   self-consistent — but #1036 already scoped "which keys every short circuit
   should populate" as a separate question.

Option 2 looks like the better shape now that there are two instances, but it
is a contract decision rather than a bug fix and should be made deliberately.

### In-boundary assessment

In boundary: `predictive_beta`'s `no_amihud_hurvich_fit` branch and the
sentence(s) describing what its inference keys mean. Out of boundary:
changing `hac_applied`'s meaning on the success path, and the general
question of short-circuit metadata completeness (that belongs with #1013's
follow-ups if it is revisited).

Ref: `factrix/metrics/predictive_beta.py` (the `no_amihud_hurvich_fit`
`_short_circuit_output` call), `docs/reference/stat-keys-by-metric.md`.

awwesomeman:
Designed as a general rule rather than a `predictive_beta` patch: #1039 is the
fourth instance in a row of the same shape (#1034, #1035, #1036), so the
remedy is the contract, not the fourth patch.

**Revised after review** — the first version granted `predictive_beta` a
documented exception on a premise that turned out to be false. See *What
changed after review* at the end.

## The rule

> A short-circuit payload may state anything true about the **input**, the
> **configuration**, or the **attempt**. A key that would describe the
> **withheld result** is neutralised — `NaN` / `None` / empty / `False` / an
> explicit failure state.
>
> The two sides are judged **per key, never per branch**.

This *describes* what the library already does:

| metric | key | short-circuit value |
|---|---|---|
| `mfe_mae` | `mfe_mae_ratio` | `float("nan")` |
| `event_around_return` | `baseline_bar_return` / `per_offset` | `None` / `{}` |
| `oos_decay` | `status` / `sign_flipped` | `"VETOED"` / `False` |
| `predictive_beta` | `stambaugh_adjusted` | `False` |
| **`predictive_beta`** | **`hac_applied`** | **`True` → `False`** |

The last two are keyword arguments to the *same* `_short_circuit_output` call:
the same claim about two different stages, decided two different ways.

A branch may instead document that a withheld-result key describes the
attempt, but only where neutralising is **impossible or destroys
information** — not merely less convenient — and must say which applies. No
metric currently needs it; the bar is written down so a future case is argued
rather than assumed.

## The envelope doc was false library-wide

`stat-keys-by-metric.md` ended its short-circuit section with:

> The auxiliary `metadata` keys listed in the per-metric subsections above are
> *not* present on the short-circuit path.

An AST sweep finds **39 call sites across 17 metric modules** passing keys
outside its allowlist (the reviewer's independent sweep, under a stricter
allowlist, found 71/23 — different line, same conclusion). Almost all are
harmless configuration echoes. The sentence described a rule the code does not
follow, and drew the line in the wrong place: the real split is
input-vs-withheld-result, not auxiliary-vs-not. Rewritten.

That is the larger half of this PR, and why the fix is not a one-liner.

## `predictive_beta`, judged per key

| key | side | value |
|---|---|---|
| `har_lags`, `newey_west_lags` | resolved for the attempt — the resolution ran | the resolved bandwidth |
| `beta_ols_uncorrected`, `factor_std` | measured before the bail | the measurement |
| `hac_applied`, `stambaugh_adjusted` | claims a stage ran | `False` |

`(hac_applied=False, har_lags=<int>)` is unreachable on any computed result.
That is the argument *for* it: an impossible combination is detectable, where
the plausible `(True, <int>)` reads as a completed fit. `hac_applied` has no
consumers outside `predictive_beta` and its own tests.

## FX008

Lints the mechanically decidable half: an `*_applied` / `*_adjusted` /
`*_flipped` key reporting a stage as having run on a short-circuit payload.
FX007's migration-ledger shape — baseline plus a staleness check.

- **The baseline is empty**, so the rule holds with no carve-outs.
- Verified to fire: planting `hac_applied=True` back into `predictive_beta`
  fails FX008 and names the site.
- The grammar matches eight names across the package, not just the one that
  prompted it.
- `*_corrected` was dropped after review: it matched `ar1_phi_corrected`, a
  float (the bias-corrected AR(1) coefficient), not a did-this-run flag, and
  legitimately attempt-side. `*_lagged` stays out too — `weights_lagged`
  echoes the `lag_weights` parameter.
- What a lint cannot decide — whether `mfe_mae_ratio` was neutralised — is
  stated in the envelope instead, and the lint's docstring says it does not
  cover that half.

## What changed after review

The first version kept `hac_applied=True` under a documented exception,
arguing that neutralising the pair would make the branch indistinguishable
from an `h = 1` success. **That premise was false.** Comparing the payloads
key by key: 27 differ, **0 share a value**. `stambaugh_adjusted` — neutralised
one keyword earlier in the same call, `True` on every computed result —
separates them on its own.

With the premise gone, the rule decides the case without an exception: the
first version had bundled `har_lags` and `hac_applied` as "the inference
keys" when the rule operates per key. `test_short_circuit_stays_distinguishable_from_an_h1_success`
pins the refutation so the premise cannot quietly return.

An earlier draft of this description cited `CLAUDE.md` as independent support
for the exception route. It is not: those sections are uncommitted and were
written this session as a codification of #1034/#1035/#1036/#1039 themselves,
so citing them was the same judgment restated. Removed.

## Validation

- `pytest -p no:faulthandler -q` — 3813 passed, 45 skipped (main: 3805; +8)
- `ruff check` / `ruff format --check` / `mypy factrix`
- `mkdocs build --strict`, no generated-file drift

One metadata **value** changes: `hac_applied` on `no_amihud_hurvich_fit`,
`True` → `False`. No estimator, statistic or p-value changes.

Closes #1039

### DIFF
--- a/docs/development/architecture.md
+++ b/docs/development/architecture.md
@@ -978,7 +978,7 @@ Hard constraints — violating these breaks the API contract:
    `False` / an explicit failure state) unless the metric documents, at that
    branch, that the key describes the attempt instead. FX008 in
    `tests/stats/test_short_circuit_flag_polarity.py` lints the decidable
-   half — a `*_applied` / `*_adjusted` / `*_corrected` / `*_flipped` key
+   half — a `*_applied` / `*_adjusted` / `*_flipped` key
    reporting a stage as having run — on the same baseline shape as FX007.
    The contract itself is
    [Which auxiliary keys a short circuit may carry](../reference/stat-keys-by-metric.md#which-auxiliary-keys-a-short-circuit-may-carry).
--- a/docs/development/architecture.md
+++ b/docs/development/architecture.md
@@ -972,6 +972,16 @@ Hard constraints — violating these breaks the API contract:
    `tests/stats/test_degenerate_guard_polarity.py` rejects new bare
    comparisons and carries the pre-existing sites as a baseline — that
    baseline is a migration ledger, not a bug list.
+   A short circuit runs no kernel, so its `metadata` may state anything true
+   about the input, the configuration or the attempt, but a key that would
+   describe the withheld result is neutralised (`NaN` / `None` / empty /
+   `False` / an explicit failure state) unless the metric documents, at that
+   branch, that the key describes the attempt instead. FX008 in
+   `tests/stats/test_short_circuit_flag_polarity.py` lints the decidable
+   half — a `*_applied` / `*_adjusted` / `*_corrected` / `*_flipped` key
+   reporting a stage as having run — on the same baseline shape as FX007.
+   The contract itself is
+   [Which auxiliary keys a short circuit may carry](../reference/stat-keys-by-metric.md#which-auxiliary-keys-a-short-circuit-may-carry).
 6. Family declaration is explicit: a screening verb's input list is one family, optionally split per bucket via `expand_over` where the API supports it. Shared resolution enforces (a) the result identity `(factor, forward_periods, *sorted(params.items()))` is unique across the input — every `params` entry joins it automatically, while `metadata` never does, (b) `expand_over` names come only from `EvaluationResult.params` (or the built-in `forward_periods`), never the factor and never a `metadata` key, (c) formal p-values are populated before procedures read them. Cross-metric BHY adds the metric label to the hypothesis identity; cross-metric partial conjunction keeps the predeclared metric count fixed under insufficient endpoints. Mixed horizons warn so the caller confirms whether selection is pooled or predeclared per horizon.
 7. A metric whose effective sample is below its own hard floor raises `InsufficientSampleError` under `strict=True` (per metric, per axis, on the effective post-stride/post-drop sample — not a global `T` rule); metrics never silently produce a result on under-sampled data. NW HAC bandwidths read `overlap_periods` and never fall below `h - 1`; scalar series means and rank-one contrasts use the wider calibrated `3(h - 1)` floor, while multi-restriction Wald paths retain the Hansen-Hodrick floor.
 
--- a/docs/reference/stat-keys-by-metric.md
+++ b/docs/reference/stat-keys-by-metric.md
@@ -936,16 +936,21 @@ which reports `R²` for the OLS regression.
     the metric short-circuits rather than substituting the biased OLS slope,
     and `metadata["reason"]` is `no_amihud_hurvich_fit`.
 
-    On that branch `hac_applied` and `har_lags` describe the covariance
-    branch and bandwidth **resolved for the attempt**, not a kernel that ran;
+    Each key on that payload is judged on its own side of the
+    [short-circuit envelope](#which-auxiliary-keys-a-short-circuit-may-carry),
+    not as one "inference" bundle:
+
+    | key | side | value |
+    |---|---|---|
+    | `har_lags`, `newey_west_lags` | resolved for the attempt | the resolved bandwidth |
+    | `beta_ols_uncorrected`, `factor_std` | measured before the bail | the measurement |
+    | `hac_applied`, `stambaugh_adjusted` | claims a stage ran | `False` |
+
     `beta_ols_uncorrected` is the uncorrected slope the metric declined to
-    publish as the headline. This is the documented exception the
-    [short-circuit envelope](#which-auxiliary-keys-a-short-circuit-may-carry)
-    allows, taken instead of neutralising: `hac_applied = False` implies
-    `har_lags = None` on every computed result, so neutralising the pair here
-    would make the branch indistinguishable from an `h = 1` success rather
-    than merely uninformative. `stambaugh_adjusted` stays `False` because it
-    describes the headline value, and there is none.
+    publish as the headline. The `(hac_applied=False, har_lags=<int>)` pair
+    that results is unreachable on any computed result — an impossible
+    combination a consumer can detect, rather than a plausible one that reads
+    as a completed fit.
 
     As the envelope says, branch on `reason` before reading these keys.
 - *warning*: `WarningCode.PERSISTENT_REGRESSOR` when the ADF p-value exceeds
@@ -1187,16 +1192,27 @@ that point is worth reporting.
 - **Keys that would describe the withheld result** are neutralised — `NaN`
   (`mfe_mae`'s `mfe_mae_ratio`), `None` (`event_around_return`'s
   `baseline_bar_return`), empty (`per_offset`), `False` (`predictive_beta`'s
-  `stambaugh_adjusted`, `oos_decay`'s `sign_flipped`), or an explicit failure
-  state (`oos_decay`'s `status = "VETOED"`) — **or** the metric documents, at
-  that branch, that they describe the attempt rather than a result.
-  `predictive_beta`'s `no_amihud_hurvich_fit` is the one branch that takes the
-  second route, because neutralising its pair would reuse a sentinel that
-  already means `h = 1`.
+  `hac_applied` and `stambaugh_adjusted`, `oos_decay`'s `sign_flipped`), or an
+  explicit failure state (`oos_decay`'s `status = "VETOED"`).
+
+The two sides are judged **per key, never per branch**.
+`predictive_beta`'s `no_amihud_hurvich_fit` is the worked example: `har_lags`
+stays affirmative because resolving the bandwidth really happened, while
+`hac_applied` — a past-tense claim that the HAC covariance was computed — goes
+to `False`. The resulting `(hac_applied=False, har_lags=<int>)` pair is
+unreachable on any computed result, and that is the point: an impossible
+combination is detectable, where a plausible one reads as a completed fit.
+
+A branch may instead document that a withheld-result key describes the
+attempt, but only where neutralising is **impossible or destroys
+information** — not where it is merely less convenient — and the branch must
+say which applies. No metric currently needs this; the bar is written down so
+a future case is argued rather than assumed.
 
 The mechanically decidable half is linted: FX008 in
-`tests/stats/test_short_circuit_flag_polarity.py` rejects a new
-`*_applied` / `*_adjusted` / `*_corrected` / `*_flipped` key that reports a
-stage as having run on a short-circuit payload, and carries the documented
-exception as a baseline. Whether a value-bearing key was neutralised is a
+`tests/stats/test_short_circuit_flag_polarity.py` rejects a
+`*_applied` / `*_adjusted` / `*_flipped` key that reports a
+stage as having run on a short-circuit payload. Its baseline of documented
+exceptions is empty, so the rule holds without carve-outs. Whether a
+value-bearing key was neutralised is a
 semantic question no lint settles; this section is the contract for it.
--- a/docs/reference/stat-keys-by-metric.md
+++ b/docs/reference/stat-keys-by-metric.md
@@ -914,10 +914,7 @@ which reports `R²` for the OLS regression.
   kernel **ran at** — the augmented design is `n_periods` rows and a
   Bartlett sum cannot use a lag it has no pair for, so on a long horizon this
   is narrower than the bandwidth resolved from the full series; `None` at
-  `h = 1`, where the corrected test is homoskedastic. One exception: on the
-  `no_amihud_hurvich_fit` short circuit no kernel runs at all, and the value
-  is the bandwidth resolved for the attempt. `None` is not reused there
-  because it already means `h = 1`;
+  `h = 1`, where the corrected test is homoskedastic;
   `newey_west_lags` stays the narrow covariance bandwidth resolved for the
   uncorrected OLS reference), `alpha` (the intercept the corrected slope implies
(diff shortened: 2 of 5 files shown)


awwesomeman:
## Design review — the rule is right, the one exception it grants is not. Please rework.

Not merging. The taxonomy and the envelope rewrite are a real improvement; the
single use of the escape hatch rests on a premise that does not hold, and fixing
that makes the PR stronger in three places at once.

### 1. The justification for the exception is factually false

The PR, the docstring, the test docstring and the FX008 baseline all rest on one
claim: neutralising `hac_applied` / `har_lags` would make the short circuit
*"indistinguishable from an `h = 1` success"*. I compared the two payloads key by
key (`n=20, h=19` short circuit against `n=240, h=1` success):

```
keys that DIFFER: 27   keys that match: 0
excluding hac_applied/har_lags, 25 keys still differ
```

They have **no key in common with an equal value**. `value` is NaN vs a float,
`reason` exists only on the short circuit, `method` is absent, `n_periods` is
`None`, and the short circuit carries 8 keys against the success path's 26.

The decisive one: **`stambaugh_adjusted` is neutralised to `False` one keyword
earlier in the same `_short_circuit_output` call**, and it is `True` on every
`h = 1` success. The collision this design was built to avoid is already
prevented by the neutralisation sitting immediately above it.

So the exception is not being taken because neutralising is impossible. It is
being taken where neutralising would have worked.

### 2. The rule as written already forbids what the PR does

The taxonomy is input / configuration / **attempt** versus **withheld result**.
Applied to the two keys separately rather than as "the inference keys":

- `har_lags` is a *resolved quantity*. "The bandwidth resolved for the attempt"
  is a true statement about work that actually happened — the resolution ran.
  Attempt side. Keeping it is right, and #1036 already decided that.
- `hac_applied` is a *past-tense assertion that a stage ran*. No HAC ran.
  Withheld-result side. The rule says neutralise.

The PR treats them as one unit and takes a single exception covering both. Split
them and the rule decides the case on its own, with no exception needed.

### 3. Recommended shape: `(False, 7)`

Neutralise `hac_applied` to `False`; keep `har_lags` with the documented
exception #1036 established. Every key then states something true.

The PR rejects this as *"a pair no computed result can produce, which then needs
its own documented exception anyway — a stranger one"*. I'd argue the opposite:
`(False, 7)` reads as "HAC was not applied; the bandwidth resolved was 7", which
is precisely what happened. That it is unreachable on a computed result is a
**feature** — it makes the branch detectable without reading `reason`. Compare
the two failure modes for a consumer that does not branch:

- `(True, 7)` — a consistent, plausible-looking payload asserting a kernel ran
  at bandwidth 7. A lie that reads as truth.
- `(False, 7)` — an impossible combination. Detectably not a computed result.

An honest impossibility beats a plausible falsehood.

### 4. It would also give FX008 teeth from day one

As it stands the lint ships with its only rule already excepted: the sole
`_FX008_BASELINE` entry is the exact case the lint was written for. Neutralise
`hac_applied` and the baseline is empty — the lint guards rather than documenting
a carve-out. I confirmed it fires correctly (emptying the baseline flags the real
site), so this is about what it is allowed to permit, not whether it works.

### 5. Answering the question you actually asked (Q1)

You asked whether the "...OR document it" clause is an escape hatch wide enough
to swallow the rule, and said you designed it while wanting a particular answer.

**Its first and only use is a case where it was not needed.** That is the
evidence — not that the clause is wrong in principle, but that on its debut it
absorbed a case the rule would have decided correctly. Keep the clause, and
tighten its condition: the exception is available where neutralising is
*impossible or destroys information*, not where it is merely less convenient.
Then state, at the branch, which of those applies. Here neither does.

One more thing, since it bears on how much weight the clause can carry: the PR
cites CLAUDE.md three times as the authority for this route. Those sections are
uncommitted, were written this session at the user's request, and are a
codification of #1034 / #1035 / #1036 / #1039 themselves. They are not
independent support — it is the same judgment restated. And what they actually
say is *"document the exception explicitly. Do not reuse a sentinel that already
means something else."* The sentinel-reuse hazard is the part that turns out not
to exist here.

### 6. Q3 — the envelope rewrite is in boundary

Agreed, and rewriting beat splitting it. #1039 is unintelligible while the
general contract states something untrue, and the old sentence drew the line on
the wrong axis. My own sweep lands at 71 call sites across 23 modules under a
minimal allowlist rather than your 39/17 under the envelope's actual one —
different allowlists, same conclusion, and the conclusion is not marginal.

### 7. Q4 — worth having, but the grammar has a false positive

You undersold it: the population is not 1. The grammar matches 9 distinct names
across the package (`kolari_pynnonen_applied`, `calendar_time_se_applied`,
`overlap_adjustment_applied`, `event_clustering_adjusted`, `mean_adjusted`, …),
so it has a real future catchment.

But `*_corrected` conflates two different things. `ar1_phi_corrected` matches the
grammar and is a **float** (`"ar1_phi_corrected": fit.phi_corrected`), not a
did-this-run flag. It is also a legitimate attempt-side value — the AR(1) fit
happens before the bail — so if any short circuit ever reports it, FX008 flags it
and forces a baseline entry for something that is not a flag at all. Either drop
`_corrected` from the grammar, or require the value to be a bool literal before
flagging.

A lint covering a quarter of its contract is worth having, provided the quarter
is decidable and the docstring says so — yours does. Just don't let it fire on
value keys.

### Verification

`3812 passed, 45 skipped` (main 3805, +7), matching your run. FX008 fires with
the baseline emptied. `hac_applied` has no consumers outside `predictive_beta`
and its own tests, so neutralising it is safe.

Rework the `hac_applied` value and the exception's condition and I'll re-review;
the envelope section, the doc rewrite and FX008's shape can stand as they are.

awwesomeman:
## Re-review at `ebc7f4c9` — design accepted. One word to fix, then I merge.

The per-key split is the right shape, and the reasoning now lives where a reader
will hit it: the code comment at the call site, the docstring, and a per-key
table in the docs all say the same thing.

### Verified independently

```
FX008 baseline entries: 0        affirmative sites: {}
planted hac_applied=True  -> FX008 FAILED as it should, naming the site
computed results producing (False, <int>): 0        # the pair really is unreachable

short circuit: hac_applied=False  har_lags=7     stambaugh_adjusted=False
h=1 success:   hac_applied=False  har_lags=None  stambaugh_adjusted=True
keys differing after the fix: 26
```

`3813 passed, 45 skipped`; `ruff check`, `ruff format --check`, `mypy factrix`
clean; CI 17 pass / 2 skipping, `CLEAN`.

Two things worth calling out as *good* rather than merely correct:

- `test_short_circuit_stays_distinguishable_from_an_h1_success` asserts the real
  invariant rather than the naive "no key shares a value" — which, as you found,
  stops being true the moment `hac_applied` is `False` on both sides. Pinning the
  refutation via `stambaugh_adjusted` and `reason` is the version that survives
  the fix it was written about.
- Bounding the exception clause and then recording that **no metric currently
  needs it** is stronger than deleting the clause would have been. The mechanism
  is there with a bar attached, and the empty baseline is the evidence the bar is
  real.

### The one fix

`_STAGE_FLAG`'s docstring says the grammar *"matches nine names across the
package today"* and then lists eight. The grammar matches eight:

```
calendar_time_se_applied     event_clustering_adjusted   hac_applied
kolari_pynnonen_applied      mean_adjusted               overlap_adjustment_applied
sign_flipped                 stambaugh_adjusted
```

Nine was the count *before* `_corrected` was dropped — `ar1_phi_corrected` was
the ninth. Your message to me said eight, so this is just the docstring lagging
the change.

Trivial, but it is a count that contradicts the list beside it, in the PR that
codifies "a message must not contradict what sits next to it". Change `nine` to
`eight` and I will merge on green.

awwesomeman:
## Re-review at `114a1dfe` — not scope creep. One line left, and it is the same hole again.

**The extra test was the right call, not scope creep.** It is a guard for the
exact defect flagged, in the file the PR already touches, and it converts a
hand-maintained count into a measured one — which is the rule this PR codifies,
applied to itself. Shipping the one word alone would have left the next person to
touch the grammar hitting the identical problem. Keep it.

It also does the primary job. Restoring `nine` fails it, as you said:

```
A: count word stale ('nine') -> FAILED
   "the grammar matches 8 names (eight); the comment on _STAGE_FLAG states a
    different count."
```

### But the second drift direction is still open

The name check is `assert name in own_source` — presence anywhere in the file,
not in the `_STAGE_FLAG` comment. Two of the eight names also appear inside
`test_fx008_detects_a_planted_violation`'s planted module string, so they can be
deleted from the list and the guard still passes:

```
drop sign_flipped from the list               -> PASSED   (still on lines 165, 173)
drop overlap_adjustment_applied from the list -> PASSED   (still on lines 165, 172)
drop mean_adjusted from the list              -> FAILED   (appears nowhere else)
```

That is the original defect reachable again: delete a listed name, the count word
stays `eight`, the comment now lists seven and claims eight, and the guard is
green. It closes the drift that happened and leaves open the one that produces
the same contradiction.

Scope the haystack to the comment block rather than the file — the `#:` lines
immediately above `_STAGE_FLAG =` — and both directions are shut.

While you are in there: the inline comment says *"every name the comment lists is
one the grammar really matches"*, but the loop asserts the converse — every name
the grammar matches appears in the source. Both directions are worth having; the
comment should describe whichever the code does.

### Everything else verified

`3814 passed, 45 skipped`; ruff + mypy clean; CI 17 pass / 2 skipping, `CLEAN`.
The word map fails loudly rather than silently when the count moves past 12,
which is the right failure mode.

This is the last item — one line, then I merge on green.

awwesomeman:
## Verified at `bd162a4a` — merging

Set equality was the better fix than the scoping I asked for. Scoping alone
closes the deletion case; set equality also closes the converse — a name listed
that the grammar no longer matches, which is how `ar1_phi_corrected` would have
been left behind had it not been caught. Both sides produce the same defect, a
list contradicting its own count, and both now fail.

All five directions checked against the real file:

```
unmodified                                       -> passed
delete sign_flipped from the list                -> FAILED  (passed before)
delete overlap_adjustment_applied from the list  -> FAILED  (passed before)
restore the stale "nine"                         -> FAILED
list a name the grammar does NOT match           -> FAILED  (ar1_phi_corrected)
make the sentence unfindable                     -> FAILED  (asserts, not silent)
```

`3814 passed, 45 skipped`; ruff + mypy clean; CI 17 pass / 2 skipping, `CLEAN`.

### On the design, now that it is settled

The shape that shipped is not the one this PR opened with, and the difference is
the point. It opened treating `hac_applied` and `har_lags` as one "inference"
bundle protected by a documented exception; it ships judging each key on its own
side of the envelope, with the exception clause bounded and unused. The rule got
narrower and the code got more honest at the same time, which is usually the sign
the rule was real rather than reverse-engineered from the case at hand.

Two things I would keep from how this went: the premise behind the first design
was checkable in about a minute, and checking it is what moved the design; and
the guard written to close the count defect had to be tested against the defect
in both directions before it could be trusted, which is the same standard the PR
is about.

Closes #1039.

PR #1041, opened 2026-09-02
ci(docs): retry the intersphinx inventory fetch before --strict fails

### LINKED ISSUE #1037: CI: strict docs build goes red when a third-party intersphinx host is unreachable
Observed on #1034's `docs-deploy-dev / build` job (run 33599419309), which failed
with no content problem in the PR:

    ERROR - mkdocstrings: Couldn't load inventory
            https://docs.pola.rs/api/python/stable/objects.inv through handler 'python':
            <urlopen error [Errno 104] Connection reset by peer>
    Aborted with 1 errors in strict mode!

A rerun went green. `mkdocs build --strict` passes locally with no
generated-file drift.

Under `--strict`, a failed intersphinx inventory fetch is a hard error, so any
PR that touches docs can go red because a third-party docs host was briefly
unreachable. That is infrastructure availability, not a defect in the change
under test, and it costs a rerun cycle each time plus the risk of a real docs
failure being read as "just the polars flake".

### Options

1. Vendor / cache the external inventories and point `mkdocstrings` at the
   local copies, refreshed on a schedule rather than per-build.
2. Keep `--strict` but downgrade an inventory *fetch* failure specifically to a
   warning, leaving every other strict-mode error fatal.
3. Retry the fetch with a short backoff before failing.

Option 1 also makes the docs build reproducible offline; option 2 is the
smallest change but needs care not to mask a genuinely wrong inventory URL.

### Note

This is distinct from the other known flake mode in this repo, where macOS
pytest jobs are CANCELLED by runner queue timeout rather than failing;
`gh run rerun <id> --failed` clears that one.

awwesomeman:
## What the code actually does today

`mkdocstrings` downloads each inventory in `mkdocs.yml` through
`mkdocs.utils.cache.download_and_cache_url`, with a **one-day cache** and
**exactly one attempt**. A failed fetch is fatal under `--strict`.

Three consequences, all verified against the installed versions
(mkdocstrings 1.0.4, mkdocs 1.6.1):

- the cache lives at `platformdirs.user_cache_dir("mkdocs")`, **not** the
  mkdocs `cache_dir` setting, and CI does not persist it — `enable-cache: true`
  on `setup-uv` caches the *uv package* cache, a different thing. So every CI
  build fetches, while a local build fetches about once a day;
- `download_and_cache_url` has **no stale fallback**: on failure
  `content = download(url)` propagates, even when a good copy is sitting in the
  cache;
- so any PR touching docs is one connection reset away from red. That is what
  happened on #1034's `docs-deploy-dev` run 33599419309.

## The change

A pre-build hook fetches the same URLs into the same cache first, retrying a
connection-level failure. It calls the *same* cache function with the *same*
one-day duration, so a warmed entry is what `mkdocstrings` finds a moment
later — it never reaches the network.

Roughly issue option 3, implemented where it also helps local builds, and
without vendoring inventories (option 1) or weakening `--strict` (option 2).

**What the retry deliberately does not do:**

- **It does not retry a `4xx`.** A 404/403/410 is a verdict about *the URL* -
  a mistyped or moved inventory - and is raised on the first attempt rather
  than hidden behind three slow tries. This is the risk the issue flagged for
  option 2. A `5xx` and `429` *are* retried: those are verdicts about the
  server's state, they are the "third-party host unreachable" case #1037 is
  about, and they cannot conceal a typo because a typo does not produce a
  `5xx`. (Revised in review - the first version retried only `URLError`, which
  left 429/502/503/504 on the single attempt this PR exists to replace.)
- **It is not fatal itself.** When every attempt fails, the hook warns and lets
  `mkdocstrings` proceed, so a genuinely unreachable or wrong URL still fails
  the build with the handler's own error. Raising in the hook would move the
  failure and lose that context while still failing. `--strict` keeps its
  meaning; only the number of chances a transient blip gets changes.

URLs are read from `mkdocs.yml` rather than repeated in the hook, so adding an
inventory warms it automatically — pinned by test.

## Every guard checked against the unfixed state

| planted defect | result |
|---|---|
| `_ATTEMPTS = 1` (no retry) | 2 tests fail |
| retry `HTTPError` too | `test_does_not_retry_a_server_verdict` fails |
| make `warm()` fatal | `test_warm_up_is_not_itself_fatal` fails |
| unregister the hook | `test_hook_is_registered_before_the_generators` fails |

**The first version of the retry test could not fail.** It derived its expected
attempt count from `_ATTEMPTS`, so setting `_ATTEMPTS = 1` still passed all
seven tests — a test reading the constant it exists to guard. It now hard-codes
the second attempt, and the table above is after that fix.

`test_cache_duration_matches_what_mkdocstrings_asks_for` reads mkdocstrings'
own source, so if upstream changes the duration the mismatch fails loudly
instead of the warm-up silently writing entries that are already stale.

## Considered, not included

Caching `~/.cache/mkdocs` in CI would cut the fetch to roughly once a day per
runner cache. It is **complementary rather than an alternative** — the retry
still covers the build that does fetch — but it touches four job definitions
across two workflows, and the retry alone closes the reported failure. Left out

Filed as #1042 with the reasoning, rather than left in this description: a
deferral that lives only in a PR body is lost the moment the PR merges.

## Validation

- `pytest -p no:faulthandler -q` — 3821 passed, 45 skipped (main: 3814; +7)
- `ruff check` / `ruff format --check` over `factrix tests scripts`
- `mypy factrix`
- `mkdocs build --strict` — hook fires first, build clean, no generated-file
  drift

No library code changes; this is build tooling and its tests.

Closes #1037

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -37,6 +37,9 @@ theme:
         name: Switch to light mode
 
 hooks:
+  # pre-warms the intersphinx inventory cache mkdocstrings reads, with
+  # retries, so a transient fetch failure cannot fail a --strict build
+  - scripts/mkdocs_hooks/warm_inventory_cache.py
   # examples/*.ipynb -> docs/examples/*.md
   - scripts/mkdocs_hooks/render_example_notebooks.py
   # factrix._metric_index.public_specs() -> docs/reference/_generated_metric_matrix.md
--- a/scripts/mkdocs_hooks/warm_inventory_cache.py
+++ b/scripts/mkdocs_hooks/warm_inventory_cache.py
@@ -59,27 +59,40 @@ def _inventory_urls(config: Any) -> list[str]:
     ]
 
 
-def _download_with_retry(url: str) -> bytes:
-    """``mkdocs``' downloader, retried on a connection-level failure.
+def _is_transient(error: Exception) -> bool:
+    """Whether the failure is worth another attempt.
 
-    Only ``URLError`` is retried. An ``HTTPError`` — 404 on a mistyped URL,
-    403, 500 — is a server verdict rather than a blip, and is raised on the
-    first attempt so a wrong URL is not hidden behind three slow tries.
+    The split is 4xx against everything else, and it is a split between two
+    kinds of statement:
+
+    - a ``4xx`` other than ``429`` is a verdict about *the URL* — ``404`` and
+      ``410`` on a mistyped or moved inventory, ``403`` on one that needs
+      credentials. Retrying only hides a wrong URL behind three slow tries;
+    - a ``5xx`` is a verdict about *the server's state*, and ``429`` asks for
+      exactly one thing, which is to come back later. A docs CDN answering
+      ``502`` or ``503`` is the third-party host being unreachable — the
+      condition #1037 is about — and cannot conceal a typo, because a typo
+      does not produce a ``5xx``.
+
+    ``HTTPError`` subclasses ``URLError``, so it is checked first.
     """
+    if isinstance(error, urllib.error.HTTPError):
+        return error.code >= 500 or error.code == 429
+    return isinstance(error, (urllib.error.URLError, OSError))
+
+
+def _download_with_retry(url: str) -> bytes:
+    """``mkdocs``' downloader, retried while the failure looks transient."""
     from mkdocstrings._internal.handlers.base import _download_url_with_gz
 
-    last: Exception | None = None
     for attempt in range(_ATTEMPTS):
         try:
             return _download_url_with_gz(url)
-        except urllib.error.HTTPError:
-            raise
         except (urllib.error.URLError, OSError) as error:
-            last = error
-            if attempt < _ATTEMPTS - 1:
-                time.sleep(_BACKOFF_SECONDS[attempt])
-    assert last is not None
-    raise last
+            if not _is_transient(error) or attempt == _ATTEMPTS - 1:
+                raise
+            time.sleep(_BACKOFF_SECONDS[attempt])
+    raise AssertionError("unreachable: the loop returns or raises")
 
 
 def warm(urls: list[str]) -> None:
--- a/scripts/mkdocs_hooks/warm_inventory_cache.py
+++ b/scripts/mkdocs_hooks/warm_inventory_cache.py
@@ -1,0 +1,110 @@
+"""Pre-warm the intersphinx inventory cache so one blip cannot fail the build.
+
+``mkdocstrings`` downloads each configured inventory through
+``mkdocs.utils.cache.download_and_cache_url`` with a one-day cache, and a
+failed fetch is fatal under ``--strict``. It is a single attempt: on
+``docs-deploy-dev`` run 33599419309 a ``Connection reset by peer`` from
+``docs.pola.rs`` failed a pull request whose content was fine, and a rerun went
+green (#1037).
+
+This hook fetches the same URLs into the same cache first, retrying a
+connection-level failure. Because it calls the same cache function, a warmed
+entry is what ``mkdocstrings`` finds a moment later — it never reaches the
+network at all. When every attempt fails the hook stays quiet about it beyond a
+warning and lets ``mkdocstrings`` proceed, so a genuinely wrong or permanently
+unreachable URL still fails the build with its own error. ``--strict`` keeps
+its meaning; only the number of chances a transient blip gets changes.
+
+The URLs are read from ``mkdocs.yml`` rather than repeated here, so adding an
+inventory there warms it automatically.
+
+Usage (manual)::
+
+    python scripts/mkdocs_hooks/warm_inventory_cache.py
+
+MkDocs hook usage (automatic, via ``hooks:`` in mkdocs.yml)::
+
+    ``on_config(config)`` runs before mkdocstrings resolves its handlers.
+"""
+
+from __future__ import annotations
+
+import datetime
+import time
+import urllib.error
+from typing import Any
+
+#: Matches the ``cache_duration`` mkdocstrings passes, so a warmed entry is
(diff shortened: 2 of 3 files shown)


awwesomeman:
## Review — the mechanism is right. Three changes, then merge.

The diagnosis is the valuable half of this PR: finding that mkdocstrings already
caches but at `user_cache_dir("mkdocs")` rather than `cache_dir`, that CI never
persists it, and that `download_and_cache_url` has no stale fallback, is what
makes "warm the same cache through the same function with the same TTL" the
right shape instead of a retry bolted on somewhere else. That part I'd not
change.

### 1. Yes — 5xx and 429 should retry. This is the one that matters

You offered the line and it is the better one. Measured against the current
hook:

```
404 Not Found            attempts=1    URL is wrong - a verdict
403 Forbidden            attempts=1    URL/auth  - a verdict
429 Too Many Requests    attempts=1    TRANSIENT - rate limit
502 Bad Gateway          attempts=1    TRANSIENT - CDN/origin
503 Service Unavailable  attempts=1    TRANSIENT - host down briefly
504 Gateway Timeout      attempts=1    TRANSIENT - upstream slow
URLError(reset)          attempts=3    TRANSIENT - the one #1037 hit
```

Four transient modes get exactly the single attempt this PR exists to replace.
That matters more than the split looks: #1037 is filed as *"CI should not go red
on a third-party docs host being unreachable"*, and a CDN answering 503 **is**
that host being unreachable. A Cloudflare-fronted docs site is at least as
likely to surface an origin problem as 502/503/52x as it is to reset the
connection — the reset is one manifestation, not the general case.

Your concern stays intact under the sharper line, because it is a 4xx concern: a
mistyped URL produces 404/403/410, which are verdicts about *the URL*. 5xx is a
verdict about *the server's current state* and cannot hide a typo. So:
`code >= 500 or code == 429` retries, everything else raises on the first
attempt.

(Ordering is right as written — `HTTPError` subclasses `URLError`, so the
`except HTTPError` clause has to come first, and it does.)

### 2. `_inventory_urls` is untested, and it can silently disarm the hook

No test references `_inventory_urls` or `on_config`. The function that decides
*what gets warmed* has no coverage, and it fails open:

```
_inventory_urls(real config)                  -> [polars, numpy]   (2 URLs)
same config with the key renamed to 'import'  -> []                 silent no-op
```

mkdocstrings has renamed that key before — `import` became `inventories`. If it
moves again, `_inventory_urls` returns `[]`, `warm()` loops over nothing,
**every test in the file still passes**, and the build quietly loses the
protection this PR adds. `warm()`'s `except Exception` widens the same hole: a
shape error in the config walk degrades to a no-op that prints nothing, because
the loop never runs.

You already applied the right instinct one function over —
`test_cache_duration_matches_what_mkdocstrings_asks_for` asserts against
mkdocstrings' own source so the TTL cannot drift silently. The config key needs
the same treatment: call `_inventory_urls(load_config("mkdocs.yml"))` and assert
it returns the non-empty set `mkdocs.yml` declares.

### 3. `test_hook_is_registered_before_the_generators` does not check "before"

Its docstring says *"It must run in the build, and first"*. The body is:

```python
assert "scripts/mkdocs_hooks/warm_inventory_cache.py" in hooks
```

Membership only — no index, no ordering. Either assert the position or rename
the test to what it checks. Same shape as the two we closed in #1040, which is
why I would rather name it than let it pass.

### On your second question — the swallow is not hiding anything

Agreed, keep it. `warm()` can only add chances; it never removes a failure,
because mkdocstrings independently retries the same URL and reports in its own
terms. `--strict` keeps its meaning for a genuinely dead host. Raising in the
hook would fail the same builds with worse context. The only leak was the
`except Exception` interaction in finding 2, and that is fixed by testing the
walk rather than by narrowing the catch.

### On the deferred CI cache — agreed, but file it

Right call to leave out: complementary, touches four job definitions across two
workflows, and the retry alone closes the reported failure. But a deferral that
lives only in a PR body is lost the moment the PR is merged. File it as its own
issue, with the reasoning already written.

### Verification

`3821 passed, 45 skipped` (main 3814, +7), matching your run. Good catch on your
own retry test deriving the attempt count from `_ATTEMPTS` — that is the third
instance of the shape this session, and hard-coding the second attempt is the
right fix. Worth noting it was found by planting the regression, not by reading
the test.

awwesomeman:
## Re-review at `d63ccc08` — all four addressed. Merging.

### The 4xx/5xx split, measured

```
code   attempts   expected
 400          1          1
 403          1          1
 404          1          1
 410          1          1
 429          3          3
 500          3          3
 502          3          3
 503          3          3
 504          3          3
reset         3          3
```

Refactoring the decision into `_is_transient` was better than threading the
condition through the loop. The split now reads as what it is — a verdict about
the URL against a verdict about the server's state — and `test_retries_a_verdict_
about_the_server` / `test_does_not_retry_a_verdict_about_the_url` name the same
distinction, so the code and the tests agree on *why* rather than only on *what*.
`test_recovers_from_a_transient_gateway_error` mirroring the reset case is the
one that pins the actual #1037 scenario in its 5xx form.

### Every guard planted against its own unfixed state

```
baseline                          18 passed
drop the 5xx/429 clause            6 failed
rename the key the walk reads      2 failed
move the hook out of position 0    1 failed
```

`test_config_walk_is_not_silently_empty_on_a_renamed_key` is the part I would
have missed. Pinning `_inventory_urls` against the real config closes the silent
`[]`; adding a companion that *renames the key and requires the result to
change* is what stops the pin from becoming a second thing that passes
vacuously. That is the failure mode we have now closed four times this week, and
this is the first time the guard-for-the-guard was written without being asked
for.

### Verification

`3832 passed, 45 skipped` (main 3814, +18); `ruff check`, `ruff format --check`,
`mypy factrix` clean; CI 17 pass / 2 skipping, `CLEAN`. #1042 filed for the CI
cache with the deferral reasoning moved out of the PR body, where it would have
died at merge.

Closes #1037.

PR #1043, opened 2026-09-02
test(docs): validate the repo-root markdown links

awwesomeman:
## Why nothing covered this

`mkdocs.yml` already sets

```yaml
validation:
  links:
    not_found: warn
    anchors: warn
```

and `--strict` turns those into errors, so every link and anchor inside
`docs/**` is checked by the build. The absence of a test was not an oversight —
it was correct for the files mkdocs builds.

The four repo-root files are outside the build. `README.md`, `CONTRIBUTING.md`,
`CLAUDE.md` and `CHANGELOG.md` are never rendered, so nothing notices when a
target moves.

## The population

Five relative links, all valid today, derived from the files rather than
listed:

| source | target |
|---|---|
| `README.md` | `LICENSE` |
| `CONTRIBUTING.md` | `docs/development/contributing.md` (×2) |
| `CLAUDE.md` | `docs/development/architecture.md#period-grid-not-calendar` |
| `CLAUDE.md` | `docs/development/contributing.md#testing-rules` |

`CLAUDE.md`'s two carry anchors, which is the failure that stays silent
longest: renaming `## Testing rules` or `## Period grid, not calendar` leaves
the link pointing at a file that still exists, at a section that no longer
does — and `CLAUDE.md` is the file agents read as authority.

Deriving the population means a link added tomorrow is covered without editing
the test. External links are left to the `link-check` workflow; reaching the
network from the unit suite would make it flaky for a reason unrelated to the
repo.

## Four planted failures, not three

| planted defect | result |
|---|---|
| rename a heading an anchor points at | 1 failure |
| point a link at a missing file | 1 failure |
| anchor that no heading produces | 1 failure |
| **regex stops matching any link** | **2 failures** |

The fourth is the one that decides whether this file is worth having. Every
resolution assertion is a loop over what the regex found, so a regex that
matches nothing turns the module green — the shape that makes a link checker
worthless, and the shape closed three times this week in #1038 and #1040.

Two assertions cover it: the scan must find links at all, and must still reach
every root file that carries one. With `_LINK` neutered the module reports
`2 failed, 1 skipped` rather than passing — the `skipped` being the
parametrised test collecting zero cases, which is exactly why the population
assertions sit outside it.

`test_at_least_one_link_is_anchored` is the same idea for the anchor branch:
it is the risky half, and it should not be possible for it to stop running
silently.

## Notes

- `_HEADING` matches ATX headings only. Setext (`===` underlines) is not used
  in this repo's markdown; if that changes, the population assertions fail
  loudly rather than the anchor check quietly missing a target.
- Anchor slugs follow GitHub's rule (lowercase, punctuation dropped, spaces
  hyphenated), since these files are read on GitHub rather than through mkdocs.
- New file rather than an addition to `tests/_doc_validation.py`, which is
  hard-scoped to `DOCS_ROOT = pathlib.Path("docs")`, or to `test_docs_pages.py`
  / `test_docs_paths.py`, which are about `docs/**` pages and package
  `docs_path` attributes. `tests/test_readme_quickstart.py` is the precedent
  for a root document owning its own test file.

## Validation

- `pytest -p no:faulthandler -q` — 3839 passed, 45 skipped (main: 3832; +7)
- `ruff check` / `ruff format --check`
- rebased onto `05259e88` and re-run before opening

Test-only; no library, docs or build-config changes.

### DIFF
--- a/tests/test_root_markdown_links.py
+++ b/tests/test_root_markdown_links.py
@@ -1,0 +1,122 @@
+"""Relative links in the repo-root markdown files resolve.
+
+``mkdocs.yml`` sets ``validation.links.not_found`` and ``.anchors`` to ``warn``,
+which ``--strict`` turns into errors, so every link inside ``docs/**`` is
+already checked by the build. The repo-root files are outside it: ``README.md``,
+``CONTRIBUTING.md``, ``CLAUDE.md`` and ``CHANGELOG.md`` are never built, so
+nothing notices when a target moves.
+
+``CLAUDE.md``'s two links carry anchors, which is the failure that stays
+silent longest: renaming ``## Testing rules`` or ``## Period grid, not
+calendar`` leaves the link pointing at a file that still exists, at a section
+that no longer does.
+
+External links are not checked here — that is the ``link-check`` workflow's
+job, and reaching the network from the unit suite would make it flaky for a
+reason unrelated to the repo.
+"""
+
+from __future__ import annotations
+
+import pathlib
+import re
+
+import pytest
+
+REPO_ROOT = pathlib.Path(".")
+
+#: Inline markdown links: ``[text](target)`` or ``[text](target "title")``.
+_LINK = re.compile(r"\[[^\]]*\]\(([^)\s]+)(?:\s+\"[^\"]*\")?\)")
+
+#: ATX headings. Setext headings are not used in this repo's markdown; the
+#: population assertions below fail loudly if that stops being true and a
+#: link starts pointing at one.
+_HEADING = re.compile(r"^#{1,6}\s+(.+?)\s*$", re.MULTILINE)
+
+_EXTERNAL = ("http://", "https://", "mailto:", "#")
+
+
+def _root_markdown() -> list[pathlib.Path]:
+    return sorted(REPO_ROOT.glob("*.md"))
+
+
+def _relative_links() -> list[tuple[pathlib.Path, str]]:
+    """``(source file, target)`` for every non-external inline link."""
+    found: list[tuple[pathlib.Path, str]] = []
+    for path in _root_markdown():
+        for match in _LINK.finditer(path.read_text(encoding="utf-8")):
+            target = match.group(1)
+            if not target.startswith(_EXTERNAL):
+                found.append((path, target))
+    return found
+
+
+def _slug(heading: str) -> str:
+    """GitHub's heading slug: lowercase, punctuation dropped, spaces hyphened."""
+    text = heading.strip().lower()
+    text = re.sub(r"[`*_]", "", text)
+    text = re.sub(r"[^\w\s-]", "", text)
+    return re.sub(r"[\s]+", "-", text).strip("-")
+
+
+def _anchors(path: pathlib.Path) -> set[str]:
+    return {
+        _slug(heading) for heading in _HEADING.findall(path.read_text(encoding="utf-8"))
+    }
+
+
+def test_the_scan_finds_the_links_it_is_meant_to_check() -> None:
+    """Without this the whole module passes by matching nothing.
+
+    Every assertion below is a loop over what the regex found, so a regex that
+    stops matching turns the file green rather than red — the failure mode
+    that makes a link checker worthless. Assert the scan has a population, and
+    that it still reaches every root file that carries a relative link.
+    """
+    assert _root_markdown(), "no markdown files at the repo root"
+
+    links = _relative_links()
+    assert links, (
+        "no relative links found in the repo-root markdown — either they were "
+        "all removed or _LINK stopped matching the syntax in use."
+    )
+
+    sources = {path.name for path, _ in links}
+    assert {"README.md", "CONTRIBUTING.md", "CLAUDE.md"} <= sources, (
+        f"the scan reached only {sorted(sources)}; a file that carries "
+        "relative links is no longer being read."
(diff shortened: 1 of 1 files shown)


awwesomeman:
## Review — the structure is right, the two matchers are not. Please rework.

The scaffolding is the best version of this we have built: population assertions
placed **outside** the parametrised test, so an empty population cannot make them
vanish with it; a separate `test_at_least_one_link_is_anchored` so the risky
branch cannot stop running quietly; and four planted failures rather than three.
The `2 failed, 1 skipped` observation is exactly why that placement matters, and
you only found it by planting it. None of that changes.

What does change is that both regexes are wrong in a way that produces **false
passes**, and I can reach them from `CLAUDE.md` today.

### 1. `_HEADING` matches `#` lines inside fenced code blocks

`docs/development/contributing.md` line 126 is a shell comment inside a fence:

```
git switch -c <type>/<short-description>

# edit and test
uv run pytest tests/test_<area>.py -v
```

`_HEADING` counts it, so `_anchors()` contains a phantom `edit-and-test`.
`architecture.md` happens to have 43 real and 43 matched; `contributing.md` has
10 real and 11 matched.

### 2. `_slug` strips underscores; GitHub keeps them

You chose GitHub's rule, and that choice is right — these files are read on
GitHub. But `_slug` does not implement it: `re.sub(r"[`*_]", "", text)` removes
`_` along with the formatting characters, while GitHub only strips the
formatting. **11 of `architecture.md`'s 43 headings diverge**, which is not a
corner case in a repo whose section titles are snake_case API names:

```
heading                                guard                   github
`expand_over` semantics                expandover-semantics    expand_over-semantics
`common_continuous` — time-series …    commoncontinuous-…      common_continuous-…
Inflation cost at low `n_assets`       …-low-nassets           …-low-n_assets
```

### Both are reachable now, in both directions

Appending two links to `CLAUDE.md` and running the guard:

```
[a](docs/development/contributing.md#edit-and-test)
[b](docs/development/architecture.md#expandover-semantics)
-> 9 passed
```

Both are broken on GitHub. The guard is green.

And the correct link is rejected:

```
[c](docs/development/architecture.md#expand_over-semantics)
-> 1 failed: "names an anchor that no heading in architecture.md produces"
```

That one **works** on GitHub. The false-alarm direction is the one I would worry
about more: the first person to link correctly into `architecture.md` gets a red
CI on a working link, and a guard that cries wolf is a guard people learn to
route around.

### Answering your two questions

**1. ATX-only is fine; the fence is the blind spot.** Your reasoning about setext
holds — this repo does not use it, and if that changed the population assertions
would fail loudly rather than the anchor check missing quietly. That is the right
shape. It just is not the failure the regex actually has, and the fence case
fails the other way: silently, and toward green.

**2. Targeting GitHub is right; the implementation is what diverges.** Root files
are read on GitHub, so GitHub's slug is the correct target and `docs/**` staying
on mkdocs' rule via `--strict` is correct too — the two rule sets differing is by
design, not drift, and I would leave that alone as you proposed. The problem is
only that `_slug` is not GitHub's rule.

### What I would change

Skip fenced blocks when collecting headings, and stop stripping `_`. Then pin the
slug rule itself with a small table — `expand_over` semantics, an `n_assets`
heading, one with backticks and punctuation — so the rule cannot silently drift
back to something that merely passes today's two links. Same reasoning as
`test_config_walk_is_not_silently_empty_on_a_renamed_key` in #1041: the guard
needs a guard when its correctness is a rule rather than a lookup.

### Verification

`3839 passed, 45 skipped`; CI 15/15, `CLEAN`. Your four planted failures all
behave as described — I reproduced them before looking for anything else.

awwesomeman:
## Re-review at `5e01f114` — both fixes verified, and you were right about the table. Merging.

### The two required fixes

```
broken-on-GitHub links appended to CLAUDE.md   -> 2 failed   (was: 9 passed)
#expand_over-semantics, correct on GitHub      -> 13 passed  (was: 1 failed)
```

Both false-pass directions closed and the false-alarm direction with them.

### On the pin table — you were right and I was wrong to reframe it

I checked your claim rather than taking it, because it was the whole basis for
keeping it:

```
drop the table AND reintroduce the [`*_] strip  ->  8 passed, 0 failed
```

So without those cases the underscore fix has **zero** coverage — neither
anchored link in the repo root contains an underscore. That makes the table the
regression test for the defect, not a guard for a guard, and the distinction you
drew — testing the bug rather than testing the test — is the right one. Under
the third criterion it is proportionate, not extra machinery. Trimming six cases
to four was the correct response to the criterion; dropping them would have been
an overcorrection.

### One note on your underscore sweep

Your conclusion holds — no heading uses `_` as emphasis, so keeping it is right
for every case that exists. My own count came out at 182 snake_case headings
rather than 196 (different counting basis; mine counts headings under `_headings`
and yours appears to count occurrences), and my one "emphatic" hit was
`` `__repr__` `` in a fossilised `docs/plans/` file — a dunder inside backticks,
which is code, not emphasis, and slugs identically either way. Raising it only
so the number on record is the one a re-run produces.

### Verification

`3844 passed, 45 skipped`; CI 15/15, `CLEAN`. `test_headings_inside_code_fences_
are_not_anchors` builds its own fixture rather than asserting against
`contributing.md`, which is right — it pins the rule, not today's file.

Closes the root-link gap left by `mkdocs.yml`'s validation covering only `docs/**`.

PR #1044, opened 2026-09-02
ci(docs): persist the mkdocs inventory cache across runs

### LINKED ISSUE #1042: ci(docs): persist the mkdocs inventory cache across CI runs
Split out of #1041 (closes #1037), which added a retry around the intersphinx
inventory fetch. This is the complementary half, deliberately left out of that
PR rather than bundled.

### What #1041 established

`mkdocstrings` caches each inventory for one day through
`mkdocs.utils.cache.download_and_cache_url`, but the cache lives at
`platformdirs.user_cache_dir("mkdocs")` — **not** the mkdocs `cache_dir`
setting — and CI does not persist it. The `enable-cache: true` on
`astral-sh/setup-uv` caches the *uv package* cache, which is a different
directory.

So a local build fetches roughly once a day, while **every CI build fetches**.
That asymmetry is why #1037 never reproduced locally.

### The change

Persist `~/.cache/mkdocs` across runs with `actions/cache`, so most CI builds
read the warmed entry instead of the network.

Four job definitions run `mkdocs build`:

- `.github/workflows/docs-deploy-dev.yml` — `build`, `build-main`
- `.github/workflows/link-check.yml` — `pr`, `scheduled`

### Why it is not a duplicate of #1041

They cover different windows and neither subsumes the other:

- the **retry** covers the build that does fetch — the first of the day, or
  any build after a cache miss;
- the **cache** removes the fetch from most builds entirely, so the retry has
  far fewer chances to be needed.

With both, a red build needs a cache miss *and* three consecutive failures.

### Not urgent

#1041 closes the reported failure on its own. This is a reduction in exposure
and a small build-time saving, worth doing if the flake recurs or if someone
is in these workflows anyway.

### In-boundary assessment

In boundary: the cache step in the four jobs, and its key. Out of boundary:
changing the cache duration (that is mkdocstrings' one-day value, which #1041's
`_CACHE_DURATION` is pinned to follow), and vendoring the inventories, which is
issue #1037's option 1 and a different trade — it would make the build fully
offline but couple the committed inventory to the pinned library versions.

awwesomeman:
The complementary half of #1041, split out at review so it would not be
bundled into a PR that closed a different problem.

## Why CI fetches every time

`mkdocstrings` already caches each intersphinx inventory for one day — but at
`platformdirs.user_cache_dir("mkdocs")`, **not** the mkdocs `cache_dir`
setting, so `setup-uv`'s `enable-cache: true` (which caches the *uv package*
cache) does not cover it.

That asymmetry is why #1037 never reproduced locally: a local build fetches
about once a day, every CI build fetches.

Restoring `~/.cache/mkdocs` takes most builds off the network entirely.
#1041's retry still covers the build that does fetch, so the two compose: a red
build now needs a cache miss **and** three consecutive failures.

## The key rotates daily — that is the design, not a detail

`actions/cache` saves only on a miss. A fixed key would be restored past
`mkdocstrings`' one-day expiry, re-downloaded anyway, and never re-saved —
identical to having no cache from day two. Rotating the key on the UTC date
matches the TTL, so the first build of a day warms and saves, and the rest of
the day reads it.

**No `restore-keys`**, unlike the lychee cache beside it in `link-check.yml`.
A restored older entry is stale by construction and re-downloaded regardless,
so falling back to one would be an inert line rather than a saving. Called out
because it is a deliberate deviation from the neighbouring style.

Applied to all four jobs that run `mkdocs build`: `docs-deploy-dev`'s `build`
and `build-main`, `link-check`'s `pr` and `scheduled`.

## The YAML cannot be executed here; its assumptions can

Both are pinned, in the file that already pins `_CACHE_DURATION` against
mkdocstrings' own source:

- `test_ci_cache_path_is_the_one_mkdocs_writes_to` asserts `~/.cache/mkdocs`
  against `platformdirs`' **Unix** implementation — the one the runners use,
  asserted rather than assumed because the dev host resolves somewhere else
  entirely. If mkdocs changes the appname this fails loudly instead of the
  cache silently covering nothing.
- `test_every_mkdocs_build_job_caches_the_inventories` walks both workflows and
  requires every job running `mkdocs build` to restore that exact path
  **before** the build.

| planted defect | result |
|---|---|
| remove the cache from one job | 1 failure |
| point it at `.cache` (the mkdocs `cache_dir` — the wrong directory) | 1 failure |
| move the restore after the build | 2 failures |

## Validation

- `pytest -p no:faulthandler -q` — 3847 passed, 45 skipped (main: 3844; +3)
- `ruff check` / `ruff format --check`
- `mkdocs build --strict` clean
- both workflows re-parsed as YAML after editing, and step order asserted
  programmatically rather than by reading the diff

No library code, docs or `mkdocs.yml` changes. This PR touches two workflow
files and one test module.

Closes #1042

### DIFF
--- a/.github/workflows/docs-deploy-dev.yml
+++ b/.github/workflows/docs-deploy-dev.yml
@@ -41,6 +41,22 @@ jobs:
           version: "0.8.x"
           enable-cache: true
 
+      # mkdocstrings caches each intersphinx inventory for one day, but at
+      # platformdirs.user_cache_dir("mkdocs") rather than the mkdocs cache_dir
+      # setting, so setup-uv's enable-cache does not cover it and every run
+      # would otherwise fetch (#1042). The key rotates daily to match that TTL:
+      # a fixed key would be restored past its expiry, re-downloaded anyway,
+      # and never re-saved, which is the same as having no cache from day two.
+      - name: Inventory cache key
+        id: inventory-cache-key
+        run: echo "day=$(date -u +%Y-%m-%d)" >> "$GITHUB_OUTPUT"
+
+      - name: Restore mkdocs inventory cache
+        uses: actions/cache@v6
+        with:
+          path: ~/.cache/mkdocs
+          key: mkdocs-inventories-${{ runner.os }}-${{ steps.inventory-cache-key.outputs.day }}
+
       - name: Install docs dependencies
         run: uv sync --frozen --python 3.12 --extra docs
 
@@ -68,6 +84,22 @@ jobs:
         with:
           version: "0.8.x"
           enable-cache: true
+
+      # mkdocstrings caches each intersphinx inventory for one day, but at
+      # platformdirs.user_cache_dir("mkdocs") rather than the mkdocs cache_dir
+      # setting, so setup-uv's enable-cache does not cover it and every run
+      # would otherwise fetch (#1042). The key rotates daily to match that TTL:
+      # a fixed key would be restored past its expiry, re-downloaded anyway,
+      # and never re-saved, which is the same as having no cache from day two.
+      - name: Inventory cache key
+        id: inventory-cache-key
+        run: echo "day=$(date -u +%Y-%m-%d)" >> "$GITHUB_OUTPUT"
+
+      - name: Restore mkdocs inventory cache
+        uses: actions/cache@v6
+        with:
+          path: ~/.cache/mkdocs
+          key: mkdocs-inventories-${{ runner.os }}-${{ steps.inventory-cache-key.outputs.day }}
 
       - name: Install docs dependencies
         run: uv sync --frozen --python 3.12 --extra docs
--- a/.github/workflows/link-check.yml
+++ b/.github/workflows/link-check.yml
@@ -33,6 +33,22 @@ jobs:
           version: "0.8.x"
           enable-cache: true
 
+      # mkdocstrings caches each intersphinx inventory for one day, but at
+      # platformdirs.user_cache_dir("mkdocs") rather than the mkdocs cache_dir
+      # setting, so setup-uv's enable-cache does not cover it and every run
+      # would otherwise fetch (#1042). The key rotates daily to match that TTL:
+      # a fixed key would be restored past its expiry, re-downloaded anyway,
+      # and never re-saved, which is the same as having no cache from day two.
+      - name: Inventory cache key
+        id: inventory-cache-key
+        run: echo "day=$(date -u +%Y-%m-%d)" >> "$GITHUB_OUTPUT"
+
+      - name: Restore mkdocs inventory cache
+        uses: actions/cache@v6
+        with:
+          path: ~/.cache/mkdocs
+          key: mkdocs-inventories-${{ runner.os }}-${{ steps.inventory-cache-key.outputs.day }}
+
       - name: Install docs dependencies
         run: uv sync --frozen --python 3.12 --extra docs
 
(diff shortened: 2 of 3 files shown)


awwesomeman:
## Review — design answers below. One **required** change, one **optional**.

### The date-rotating key is right, and so is dropping `restore-keys`

I worked through `actions/cache`'s save semantics rather than taking the
reasoning: it saves only on a miss, so a fixed key would be restored past
mkdocstrings' one-day expiry, re-downloaded, and never re-saved — the entry
would freeze at day-one content and every later build would fetch. Identical to
no cache from day two, exactly as you describe.

Rotating on the UTC date makes the cache lifetime equal the TTL, which is the
thing you wanted to express. The alternative — `cache/restore` plus `cache/save`
on a fixed key — would re-upload on every run to achieve the same freshness, so
it is strictly worse. I would not change it.

Dropping `restore-keys` is correct for the same reason and not an omission: a
fallback entry is stale by construction and re-downloaded regardless, so the
line would be inert. Right call to flag it, and the comment earns its place
because it sits next to the lychee cache that does use them.

### Required — `test_ci_cache_path_is_the_one_mkdocs_writes_to` does not test that

The docstring says it prevents *"if mkdocs changes the appname the cache step
silently stops covering anything"*. The assertion is:

```python
linux_path = Unix(appname="mkdocs").user_cache_dir
```

The `"mkdocs"` is written in the test, not read from the library. So it pins
**platformdirs' Unix layout for a literal string**, not the appname
`mkdocs_get_deps` actually passes. If that library moved to
`user_cache_dir("mkdocs-get-deps")`, this test still passes and the CI cache
covers nothing — the precise failure the docstring claims it catches.

For the record, the real path is:

```
mkdocs.utils.cache.download_and_cache_url
  -> mkdocs_get_deps.cache.download_and_cache_url
  -> os.path.join(platformdirs.user_cache_dir("mkdocs"), "mkdocs_url_cache")
```

so `~/.cache/mkdocs` is correct today. Read the appname out of
`inspect.getsource(mkdocs_get_deps.cache)` and assert on that — the technique
you already use two tests up in `test_cache_duration_matches_what_mkdocstrings_
asks_for`. Choosing `Unix(...)` explicitly because this dev host is Windows was
the right instinct; it was just applied to the half that could not drift.

### Optional — the runner OS is the other half of the same assumption

`~/.cache/mkdocs` is only the right path because every building job is Linux.
All four are `ubuntu-latest` today, and nothing pins it:

```
flip one building job to macos-latest  ->  21 passed
```

On macOS the directory is `~/Library/Caches/mkdocs`, so the step would cover
nothing. It is one line inside the loop you already walk over those jobs.

I am marking this optional rather than required because the scenario — moving a
docs build to macOS — is considerably less likely than a library rename, and I
would rather not spend your cycles on machinery the problem does not warrant.
If you think the one line is cheaper than remembering the assumption, take it;
if not, ship without and I will not raise it again.

### Verification

`3847 passed, 45 skipped` (main 3844, +3); CI 16 pass / 1 skipping, `CLEAN`. All
four building jobs confirmed `ubuntu-latest` and confirmed to run `mkdocs build`.
`actions/cache@v6` matches the version already used beside it in
`link-check.yml`, so the style is consistent.

Your three planted defects behave as described. The `.cache` one is the good
catch — it is a real mkdocs setting and would read as plausible in review, which
is exactly the defect a human eye passes over.

On the `git checkout` note: worth saying, and the reason it cost nothing is that
you had already written the test, so the mistake surfaced immediately instead of
silently. That is the argument for writing it first, not a near miss.

awwesomeman:
## Re-review at `000dab1f` — both items verified. Merging.

### Required item: verified against the failure, not the description

I patched `mkdocs_get_deps.cache`'s own source in the venv and ran the guard,
restoring it afterwards and checking the file back byte-identical by hash:

```
baseline                        1 passed
appname renamed upstream        1 failed
a second user_cache_dir call    1 failed
restored byte-identical: True
```

Both halves now come from the library: the delegation assertion, then the
appname pulled out of `inspect.getsource`. The `len(appnames) == 1` guard is the
part I would not have asked for and would have been wrong not to — a restructure
that adds a second `user_cache_dir` call would otherwise let the regex pick
silently, which is the same class of failure one layer down.

### Optional item: taken, and it fires from either workflow

```
docs-deploy-dev -> macos-latest    1 failed
link-check      -> macos-latest    1 failed
```

Both halves of the assumption are closed now — the appname the library passes,
and the OS that decides how it resolves.

### Verification

`3847 passed, 45 skipped`. CI counted rather than filtered, per your own note
below: **17 rollup entries, 16 pass / 1 skipping**, `mergeStateStatus CLEAN`.

### On your process note — it applies to me, and I have changed how I check

That is a real hazard and worth more than a footnote. `CLEAN` with an empty
rollup and `CLEAN` with 17 passing checks are the same value in that field, and
"no workflow ran" is indistinguishable from "everything passed" if you only read
the status. My own merge gate had the same weakness from the other direction: I
was filtering `gh pr checks` output for anything that was not `pass` or
`skipping`, which reports success just as happily when the list is empty.

So I counted the rollup on this PR before merging, and will keep doing that. You
found it in your workflow and it fixed mine — worth recording for that reason
rather than as a near-miss.

Closes #1042, and with it the series.

PR #1047, opened 2026-09-03
fix(metrics): screen the bandwidth the kernel ran on (#1045) (#1046)

### LINKED ISSUE #1045: fix(metrics): predictive_beta 的 ill-conditioned bandwidth 篩檢從不讀 kernel 實際跑的設計
Found reviewing the post-`v0.31.0` range (#1034 / #1036 / #1038 / #1040).

#1035 fixed the *name* the ill-conditioned-bandwidth warning prints, and #1034
fixed the *value* `metadata["har_lags"]` reports. Both left the screen's own
firing set alone, and that is where the same defect class survives one step
further out: the screen reads `(n_periods_finite, resolved_har_lags)`, but the
Bartlett kernel runs on `(n_periods, har_lags)` — the augmented design's
`m = n - h` rows at the clipped bandwidth.

```python
if hac_applied and _hac_bandwidth_ill_conditioned(n, resolved_har_lags):
```

The code documents the calibration choice, and the `narrowed` clause added by
#1038 says both quantities **when the bandwidth was clipped**. Where the
bandwidth was *not* clipped and only `n` and `m` diverge, nothing is said and
nothing fires — the kernel forms its lag products on a design that is thin by
the metric's own `_MIN_PERIODS_PER_LAG = 5` rule, and the caller is not told.

### Reachable

Enumerated over every `(n, h)` the metric admits (`m = n - h >= 5`, `n` 10..500),
counting cells where `_hac_bandwidth_ill_conditioned(m, har_lags)` holds but
`_hac_bandwidth_ill_conditioned(n, resolved_har_lags)` does not:

| bandwidth source | cells enumerated | thin on the design, screen silent |
|---|---|---|
| default (`newey_west_lags=None`) | 122,259 | **562** (0.46%) |
| every explicit `newey_west_lags` | 205,026 | **4,409** (2.2%) |

Examples on the default bandwidth:

| n | h | design rows `m` | `har_lags` L | rows the 5-per-lag rule wants | screen reads |
|---|---|---|---|---|---|
| 45 | 2 | 43 | 9 | 45 | `45 < 45` → silent |
| 45 | 3 | 42 | 9 | 45 | `45 < 45` → silent |
| 46 | 2 | 44 | 9 | 45 | `46 < 45` → silent |

All of them sit within one period of the boundary, so the practical cost is
small; the contract cost is not — this is the screen reading a quantity the
kernel did not run on, which is exactly what #1034–#1039 were about.

### Options

1. Screen the applied pair as well and keep the finite-pair screen —
   `_hac_bandwidth_ill_conditioned(n, resolved) or
   _hac_bandwidth_ill_conditioned(fit.n_used, fit.lags_used)` — with the
   message naming whichever comparison fired. `n_periods` and `har_lags` are
   both already in `metadata`, so the second clause looks up cleanly.
2. Replace the finite-pair screen with the applied one. **Wrong**:
   `_hac_bandwidth_ill_conditioned` gates on `n >= MIN_PERIODS_WARN = 30`, so a
   design under 30 rows would go silent — that silences the long-horizon cells
   the screen matters most on.
3. Document the gap and leave the firing set alone.

Option 1 is the one that satisfies "report what ran" without losing coverage.
It widens the warning set by the 562 default cells above.

### In-boundary assessment

In boundary: the warning's firing condition and its text. A caller reading the
warning is deciding whether to trust the p-value, and the long-run variance
they are being asked about is the one computed on the augmented design.

Out of boundary: re-calibrating `_MIN_PERIODS_PER_LAG` itself, and the parked
uneven-grid HAC calibration question in #862.

Ref: `factrix/metrics/predictive_beta.py`, the
`WarningCode.HAC_BANDWIDTH_ILL_CONDITIONED` block.

### LINKED ISSUE #1046: test(stats): _amihud_hurvich_beta 的 clipped-bandwidth df 改動是空轉且無 guard
Found reviewing the post-`v0.31.0` range (#1034).

#1034 bound the applied bandwidth in `_amihud_hurvich_beta` and routed it to
two places: the kernel call (a no-op, `_ols_nw_multivariate_from_finite`
already clips to `n - 1` internally) and the effective degrees of freedom:

```python
# The df must read the bandwidth the kernel ran at, not the wider one requested.
dof = float(max(m - 3, 1)) if h == 1 else _har_dof(m, lags_used, h)
```

That comment reads as if a batch of p-values was being corrected. **None was.**
Reverting `lags_used` to `lags` on this line leaves `tests/stats` at
`502 passed`, zero failures.

### Measured

Enumerated every `(n, h, newey_west_lags)` triple the metric admits (`n`
10..300, `m = n - h >= 5`, every explicit bandwidth plus the default):

| | count |
|---|---|
| triples enumerated | 205,026 |
| triples where the bandwidth is clipped (`lags_used < lags`) | 13,968 |
| triples where `_har_dof` differs | **0** |

The mechanism: a clip means `L >= m`, so the requested side gives
`1.5m/L - 1 <= 0.5`, and the applied side gives
`1.5m/(m-1) - 1 = (0.5m + 1)/(m - 1) < 1` for every `m >= 5`. Both land under
`_har_dof`'s floor of `1.0` and come out equal. The `T/h - 1` overlap cap is
below `1.0` in most of these cells too.

So the line is correct code that is currently inert. It is also **unguarded**:
no test fails when it is reverted, and none can, because the difference is not
observable through the public API today. It would become observable if the
`1.5` LLSW factor or the `1.0` floor ever moved.

### Asked for

1. A pinning test, labelled as one per
   [Testing rules](https://github.com/awwesomeman/factrix/blob/main/docs/development/contributing.md#testing-rules)
   — it fixes a decision against silent drift, it is not regression proof for a
   defect. It should assert the enumeration above (df identical on every
   clipped cell) so that a future change to the calibration constants surfaces
   as a failure rather than as silent p-value drift.
2. The comment corrected to say what is true: the df must read the applied
   bandwidth, *and* under today's floor the two agree everywhere, so no
   published p-value moved.

### In-boundary assessment

In boundary, but low priority — nothing is wrong with the number today. It is
filed because a comment that overstates its own effect is the exact failure the
"measure the claim; do not assert it" convention exists to catch, and because
the pin is what makes the claim survive a calibration change.

Not `parked`: there is no external trigger, it is simply small.

Ref: `factrix/_stats/ols.py`, `_amihud_hurvich_beta`'s `dof` line.

awwesomeman:
Closes #1045. Closes #1046.

Both come out of reviewing the post-`v0.31.0` range (#1034 / #1038 / #1040),
and both are the same defect class those PRs were about: a statement that is
not the quantity the code ran on.

## #1045 — the ill-conditioned screen never read the design the kernel ran on

#1034 made `metadata["har_lags"]` report the applied bandwidth. The screen
itself kept reading `(n_periods_finite, resolved_har_lags)`, while the Bartlett
sum forms its lag products on the augmented design's `n_periods` rows at the
applied bandwidth.

Enumerated over every reachable `(n_periods_finite, overlap_periods,
newey_west_lags)` cell, `n` 10..500, `m = n - h >= 5`:

| bandwidth source | cells | thin on the design, screen silent | **and no other warning fires** |
|---|---|---|---|
| default (`newey_west_lags=None`) | 122,259 | 562 | **0** |
| every explicit `newey_west_lags` | 757,735 | 18,678 | **4,252** |

The premise had to be measured before the fix: the docstring claimed the
short-periods screen owns the regime this one skips, and on the default
bandwidth that holds — all 562 cells already carry
`UNRELIABLE_SE_SHORT_PERIODS`. The hole is real only on an explicit bandwidth.
Live on `main` at c56a1e9b:

```
predictive_beta(..., overlap_periods=2, newey_west_lags=13)   # n = 65 finite pairs
  n_periods=63  n_periods_finite=65  har_lags=13   (5-per-lag wants 65 rows)
  warnings: overlapping_predictive_inference — and nothing else
```

The screen now reads both pairs and the message runs one clause per comparison
that fired, each naming the metadata key whose value it prints —
`n_periods_finite` for the finite pairs, `n_periods` for the design. Cost, so
it is on the record: the 562 default cells gain a second warning beside the one
they already carry.

The docstring sentence "shorter samples use `UNRELIABLE_SE_SHORT_PERIODS`
instead" was already false — both screens fire together today on an explicit
bandwidth — and is corrected rather than preserved.

## #1046 — the df half of #1034 is inert, and said otherwise

`_har_dof(m, lags_used, h)` carried the comment "the df must read the bandwidth
the kernel ran at", which reads as a correction to a batch of p-values. None
moved. Reverting `lags_used` to `lags` leaves `tests/stats` at 502 passed.

205,026 triples enumerated, 13,968 clipped, **0** with a different df — because
two guards equalise the sides independently:

| quantity | max over the clipped cells |
|---|---|
| `1.5m/L - 1` (requested) | 0.500 |
| `1.5m/L - 1` (applied) | 0.875 |
| `m/h - 1` (overlap cap) | **-0.375** |

A clip needs `h / n` around two thirds and up, and the cap is negative
throughout that band, so it wins the `min()` and the floor returns `1.0` either
way; strip the cap and the LLSW term is still under the floor on both sides.
The first draft of this claim blamed the floor alone and was wrong — mutating
the floor to `0.25` changes nothing, which is how it was caught.

The comment now states the measured scope, and a pinning test — labelled as a
pin, not as regression proof — holds the invariant.

## Verification

- The #1045 test fails on the unfixed code at the right assertion (the premise
  assertions above it pass, so the cell really is uncovered).
- The pin fires only when the cap **and** the floor are both loosened
  (`0.875 = _har_dof(5, 4, 8)` against `0.5 = _har_dof(5, 5, 8)`); mutating
  either alone leaves it passing, correctly, and the docstring says so.
- A false-positive guard covers the new clause: `n=240, h=5` on the default
  bandwidth stays quiet.
- `assert clipped > 1000` keeps the enumeration from passing vacuously.
- Full suite `3850 passed, 45 skipped`; `ruff check` / `ruff format --check`
  clean; `mypy factrix` clean; `mkdocs build --strict` builds.

The push gate was run manually (`--no-verify` on the push itself): the hook
shells out to `uv run mkdocs`, which is not resolvable from a worktree without
its own synced environment. Both gate commands were run against this branch
with the main checkout's interpreter and both pass.

### DIFF
--- a/docs/reference/stat-keys-by-metric.md
+++ b/docs/reference/stat-keys-by-metric.md
@@ -979,9 +979,17 @@ which reports `R²` for the OLS regression.
   flipped the verdict on draws where the two slopes are far apart.
 - *warning*: `WarningCode.UNRELIABLE_SE_SHORT_PERIODS` when
   `n_periods_effective` is below `MIN_PERIODS_WARN`.
-- *warning*: `WarningCode.HAC_BANDWIDTH_ILL_CONDITIONED` when the scalar HAR
-  bandwidth satisfies `L > T / 5` on at least 30 finite periods. Below 30,
-  `UNRELIABLE_SE_SHORT_PERIODS` owns the regime without a duplicate warning.
+- *warning*: `WarningCode.HAC_BANDWIDTH_ILL_CONDITIONED` when **either** the
+  resolved bandwidth satisfies `L > n_periods_finite / 5` on at least 30 finite
+  pairs, **or** the applied bandwidth satisfies `L > n_periods / 5` on at least
+  30 rows of the augmented design. The second pair is the one the Bartlett sum
+  formed its lag products from: an explicit `newey_west_lags` can leave the
+  design thin while the finite pairs are not, and 4,252 of the 757,735
+  reachable `(n_periods_finite, overlap_periods, newey_west_lags)` cells carry
+  no other warning when it does. The message runs one clause per comparison
+  that fired, each naming the metadata key whose value it prints. This screen
+  and `UNRELIABLE_SE_SHORT_PERIODS` are not a partition — both fire wherever
+  both hold.
 - *degenerate*: zero factor variance returns `value=NaN` with
   `WarningCode.DEGENERATE_VARIANCE` and no test.
 - *short-circuit*: `reason` `insufficient_predictive_periods`,
--- a/factrix/_stats/ols.py
+++ b/factrix/_stats/ols.py
@@ -581,8 +581,20 @@ def _amihud_hurvich_beta(
     # augmented SE is a Bartlett HAC one, so the t is read against the same
     # fixed-b effective df the scalar HAR path uses (``_har_dof``). This is a
     # single-restriction slope test, so the K x K Wald argument that keeps the
-    # multivariate paths on the narrow rule does not apply to it. The df must
-    # read the bandwidth the kernel ran at, not the wider one requested.
+    # multivariate paths on the narrow rule does not apply to it. The df reads
+    # the bandwidth the kernel ran at, not the wider one requested - though
+    # under today's calibration that moves no p-value at all, and the comment
+    # says so rather than implying a fix. Measured over every reachable
+    # (n, h, lags) triple (205026 enumerated, 13968 clipped): 0 cells where the
+    # df differs, because TWO guards independently equalise them.
+    #   - A clip needs L > m - 1, which only happens at h/n around 2/3 and up,
+    #     and there the T/h - 1 overlap cap is at most -0.375 - negative, so it
+    #     wins the min() and the floor returns 1.0 on both sides.
+    #   - Even without the cap, 1.5m/L - 1 is at most 0.5 requested and 0.875
+    #     applied, both under the same floor.
+    # So the line only starts to matter if the cap and the floor are BOTH
+    # loosened; tests/stats/test_amihud_hurvich_bandwidth_report.py pins the
+    # invariant against exactly that (#1046).
     dof = float(max(m - 3, 1)) if h == 1 else _har_dof(m, lags_used, h)
     p_value = float(2 * sp_stats.t.sf(abs(t_stat), df=dof))
     return AmihudHurvichFit(
--- a/factrix/metrics/predictive_beta.py
+++ b/factrix/metrics/predictive_beta.py
@@ -459,10 +459,14 @@ def predictive_beta(
                 "of the augmented design"
             )
         # Where the kernel ran narrower than the resolved bandwidth, say so,
-        # so the text cannot contradict metadata["har_lags"].
+        # so the text cannot contradict metadata["har_lags"] (#1038). The
+        # applied clause already prints ``fit.lags_used``, which IS
+        # ``har_lags``, so when it fires the contradiction this tail guards
+        # cannot arise and the tail would only repeat two facts already in
+        # the sentence.
         narrowed = (
             ""
-            if fit.lags_used == resolved_har_lags
+            if fit.lags_used == resolved_har_lags or applied_thin
             else (
                 f" The augmented design admits only {fit.n_used - 1} lags, "
                 f"so the kernel ran at L={fit.lags_used}."
--- a/factrix/metrics/predictive_beta.py
+++ b/factrix/metrics/predictive_beta.py
@@ -137,11 +137,16 @@ def predictive_beta(
     bandwidth and fixed-$b$ effective df this test now uses. Read an
     ``h > 1`` predictive $p$ against a raised hurdle regardless of the
     correction. ``WarningCode.OVERLAPPING_PREDICTIVE_INFERENCE`` makes that
-    known regime explicit on every such result. When the resolved bandwidth
-    also exceeds one fifth of a sample of at least 30 periods,
+    known regime explicit on every such result.
     ``WarningCode.HAC_BANDWIDTH_ILL_CONDITIONED`` records the separate kernel
-    estimation problem; shorter samples use
-    ``WarningCode.UNRELIABLE_SE_SHORT_PERIODS`` instead.
+    estimation problem, and it reads **both** pairs: the resolved bandwidth
+    against the finite pairs, which is the quantity it was calibrated on, and
+    the applied bandwidth against the augmented design's rows, which is where
+    the Bartlett sum actually formed its lag products. The two are not
+    interchangeable — an explicit ``newey_west_lags`` can leave the design thin
+    while the finite pairs are not — and they are not a partition with
+    ``WarningCode.UNRELIABLE_SE_SHORT_PERIODS`` either: both fire together
+    wherever both hold.
 
     The covariance follows the same horizon split. At
     ``overlap_periods = 1`` the corrected test uses Amihud-Hurvich's
@@ -170,10 +175,11 @@ def predictive_beta(
 
     Branch on ``metadata["reason"]`` before reading any of them.
 
-    The ``HAC_BANDWIDTH_ILL_CONDITIONED`` message names
-    ``n_periods_finite``, not ``n_periods``. This metric is the one place the
-    two diverge — the screen is calibrated on the finite pairs, while
-    ``n_periods`` counts the truncated augmented design.
+    Each clause of the ``HAC_BANDWIDTH_ILL_CONDITIONED`` message names the key
+    whose value it prints — ``n_periods_finite`` for the finite pairs,
+    ``n_periods`` for the augmented design's rows. This metric is the one place
+    the two diverge, which is why it cannot borrow the single ``n_periods``
+    token the other HAR consumers share.
 
     **The correction costs power** where OLS's apparent power was partly
     its own bias: at $T=60,\ \phi=0.95,\ \rho=-0.9$ the corrected test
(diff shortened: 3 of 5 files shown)


awwesomeman:
## Review — correct, and not over-design. One **optional** item, then merge.

Read both correction comments before the bodies, as advised. Verified in my own
worktree at `2130f153`.

### Is the OR over-design? No — it is the screen reading what the kernel ran on

I judged it against the three axes rather than by feel.

**Correctness.** The screen's job is to say the long-run variance rests on too
few lag products. The Bartlett sum forms those products on
`(n_periods, har_lags)`. Reading only `(n_periods_finite, resolved_har_lags)`
was the screen measuring a different sample than the one the kernel used — the
same defect as #1034/#1035/#1036, one layer further out. Adding the OR is not
new machinery; it is the existing screen reading the right pair.

**Style consistency.** `applied_thin` goes through
`_hac_bandwidth_ill_conditioned`, so it inherits the `n >= MIN_PERIODS_WARN`
floor. That floor is the mechanism the house convention relies on to keep the
two diagnostics non-redundant (`fm_beta`: *"below 30 periods the binding problem
is already the short score series, not a second bandwidth echo"*). The new term
does not bypass it.

**Proportionality.** The cost is bounded and you measured it. I confirmed the
premise behind the docstring correction rather than taking it — on `main` at
`c56a1e9b`:

```
n=90 h=63:  bandwidth=True  short_periods=True   -> both fire together
n=90 h=60:  bandwidth=True  short_periods=True   -> both fire together
```

So the "partition" sentence was already false before this PR, and correcting it
is not a consequence of the change. And the 562-cell case is real: `n=45, h=2`
on `main` carries `overlapping` and `short_periods` only, with 43 design rows
under a 9-lag Bartlett sum.

### The composed message, all three forms

```
both clauses (n=90 h=60):
  the resolved Bartlett bandwidth L=30 exceeds n_periods_finite / 5 on the 90
  finite pairs and the applied Bartlett bandwidth L=29 exceeds n_periods / 5 on
  the 30 rows of the augmented design, so the long-run variance rests on too few
  lag products to be stable. The augmented design admits only 29 lags, so the
  kernel ran at L=29.

applied only (n=65 h=2 lags=13):
  the applied Bartlett bandwidth L=13 exceeds n_periods / 5 on the 63 rows of
  the augmented design, so ...

resolved only (n=90 h=63):
  the resolved Bartlett bandwidth L=30 exceeds n_periods_finite / 5 on the 90
  finite pairs, so ... The augmented design admits only 26 lags, so the kernel
  ran at L=26.
```

No #1035 defect: each clause's printed count is the value under the key it
names — `n_periods_finite=90` against "the 90 finite pairs",
`n_periods=63` against "the 63 rows". Checked against the metadata in each cell.

### Optional — the `narrowed` tail is redundant exactly when the applied clause fires

You asked whether the composition reads well in both-clause cells. It does,
except at the end. In the first example above the applied clause has already
said `L=29` on 30 augmented rows, and the tail then says *"The augmented design
admits only 29 lags, so the kernel ran at L=29."* — the same two facts again.

The tail exists (from #1038) so the text cannot contradict
`metadata["har_lags"]` when only the resolved clause prints. When `applied_thin`
fires, the applied clause already prints `fit.lags_used`, which *is*
`har_lags` — so the contradiction it guards against cannot arise, and the
sentence is pure repetition. It is reachable whenever a clip coincides with
`n_used >= 30`, e.g. `n=90, h=60`.

One condition (`and not applied_thin`) closes it. **Optional** because nothing
it says is untrue — this is message quality, not correctness. Your call: push it
and I merge, or say no and I merge as-is. I will not raise it again either way.

### The pin is honest — keep it

You asked whether it reads as coverage it does not have. It states its own scope
in the docstring, and I verified that scope by mutation rather than trusting the
report:

```
floor 1.0 -> 0.25 only      1 passed
overlap cap removed only    1 passed
BOTH loosened               1 failed
```

Exactly as documented. A pin whose docstring says "this only fires if X and Y
both move" and which demonstrably only fires then is not misleading — it is
labelled correctly. It runs in 0.90s and carries `assert clipped > 1000`, so it
cannot pass by enumerating nothing. I would not delete it.

### Your requested mutation

Dropping the `applied_thin` term from the condition:

```
assert bandwidth, "the design the kernel ran on is thin and must be reported"
E   AssertionError: assert []
```

Fails at that line, not earlier — so the three cell-validity assertions above it
hold, and the cell proves what the test claims.

### Gates, re-run because of `--no-verify`

`3850 passed, 45 skipped`; `ruff check` and `ruff format --check` clean; `mypy
factrix` clean; `mkdocs build --strict` builds with no generated-file drift. CI
counted rather than filtered: **19 rollup entries, 17 pass / 2 skipping**,
`mergeStateStatus CLEAN`.

awwesomeman:
## Re-review at `0082e7a9` — verified. Merging.

Rendered all three forms against the branch (printing `factrix.__file__` first,
per your note — it resolves to the worktree):

```
both clauses + clip (n=90 h=60)   tail present: False
  ... and the applied Bartlett bandwidth L=29 exceeds n_periods / 5 on the 30
  rows of the augmented design, so ...

resolved only (n=90 h=63)         tail present: True
  ... on the 90 finite pairs, so ... The augmented design admits only 26 lags,
  so the kernel ran at L=26.

applied only (n=65 h=2 L=13)      tail present: False
```

The gate is exactly right in the direction that matters: in the resolved-only
cell the message names `L=30` while `metadata["har_lags"]` is `26`, so the tail
is the only thing preventing the #1038 contradiction — and it is kept. In the
both-clause cell the applied clause prints `L=29`, which *is* `har_lags`, so
dropping the tail removes repetition without reopening anything.

Restoring the tail unconditionally fails the new test — `1 failed, 9 passed` —
so it is not a test that cannot fail.

### Gates

`3851 passed, 45 skipped`; `ruff check` and `ruff format --check` clean; `mypy
factrix` clean; `mkdocs build --strict` builds with no generated-file drift. CI
counted: **19 rollup entries, 17 pass / 2 skipping**, `mergeStateStatus CLEAN`,
head `0082e7a9` — the commit I verified.

### On your `sys.path` correction

Worth passing back rather than just noting: the reason it bit is that a
scratchpad script puts the scratchpad on `sys.path[0]`, and an editable install
then answers `import factrix` from the main checkout, so the script silently
tests `main` while appearing to test the branch. It fails *toward* looking
broken here, which is the lucky direction — the same mistake on a passing
assertion would have read as the fix working. Printing `factrix.__file__` first
is the cheap check, and I ran it before trusting any of the output above.

`#1048` is the right call on the hook, and framing it as the repo's own
"report what ran" rule turned on its own tooling is exactly what it is.

Closes #1045 and #1046.

PR #1049, opened 2026-09-03
chore(hooks): report a gate that could not run as not run (#1048)

### LINKED ISSUE #1048: chore(hooks): pre-push gate 從 worktree 跑不起來，且把「跑不起來」報成「沒通過」
發現來源：#1047 review（factrix-25 提出，兩個 session 都撞到）。

`scripts/hooks/pre-push` 第 132 行跑 `uv run mkdocs build --strict`，第 mypy
段同樣走 `uv run`。從 **git worktree** push 時 `uv run` 找不到直譯器：

```
$ git push -u origin fix/1045-screen-reads-the-design
CHANGELOG polish + mkdocs --strict + mypy................................Failed
pre-push blocked: mkdocs build --strict failed.

$ uv run mkdocs build --strict          # 在 worktree 內重跑
error: Failed to spawn: `mkdocs`
  Caused by: program not found
```

`uv run` 找的是專案目錄下的 `.venv`，只有主 checkout 有；worktree 沒有。

兩個問題疊在一起：

1. **訊息說謊。** 它印 "mkdocs build --strict failed"，但 mkdocs 根本沒跑。
   用主 checkout 的直譯器對同一個 branch 跑 `mkdocs build --strict` 是成功的
   （26s，無 error）。一個把「工具不存在」報成「gate 不過」的 gate，會讓人去
   debug 不存在的 docs 問題。這與 repo 既有的
   [Report what ran, not what was asked for](../blob/main/CLAUDE.md) 同一類。
2. **後果是養成 `--no-verify` 習慣。** worktree-based 分支每次都會撞到，而
   繞過的方式正好是整個關掉四個 gate。#1047 就是這樣 push 的（手動跑完四個
   gate 並在 PR 內文寫明，reviewer 也重跑過，但這不該是常態）。

### 可能做法

1. hook 偵測 `uv run` 是否可用，不可用時 fallback 到 `python -m mkdocs` /
   `python -m mypy`，或直接用 `$(git -C "$(git rev-parse --git-common-dir)/.." ...)`
   指向主 checkout 的 `.venv`。
2. 至少把失敗訊息分成兩種：gate 不過 vs. gate 跑不起來，後者要說清楚是環境
   問題並給出 `uv sync --extra docs` 的指示。
3. 在 contributing 文件寫明 worktree 需要自己的 env。

(2) 是最小且立刻消除誤導的一步；(1) 才真正讓 worktree 分支不必 `--no-verify`。

### In-boundary assessment

在邊界內，但屬於 **contributor tooling**，不是 library 行為——不會產生錯誤的
數字，也不影響任何使用者可見的 API。優先度低，如實說明而不掛 `parked`：沒有
外部觸發條件，只是小。

不過它的失效模式會誘導繞過真正的 gate，所以不是純美化。

Ref: `scripts/hooks/pre-push`、`.pre-commit-config.yaml` 的 `push-gates`。

awwesomeman:
Closes #1048.

The pre-push gates shell out to `uv run mkdocs` / `uv run mypy`. `uv run`
resolves the environment of the directory it is invoked from, and a git
worktree has none:

```
$ uv run mkdocs build --strict          # inside a worktree
error: Failed to spawn: `mkdocs`
  Caused by: program not found
```

The hook reported that as `pre-push blocked: mkdocs build --strict failed`.
mkdocs never started. It is this repo's own "report what ran, not what was
asked for" rule inverted, on its own tooling — and the practical cost is worse
than the confusion, because the way out is `--no-verify`, which drops every
other gate along with the one that could not run. #1047 was pushed that way.

## What changed

The runner is resolved **before** the gate runs:

1. `uv run` when it can actually start the tool — the normal main-checkout
   case, so the auto-sync against the lock is kept;
2. otherwise the main checkout's `.venv`, which a worktree shares a `.git`
   with. This is the part that makes a worktree push *work*, not merely fail
   more clearly;
3. otherwise the gate is reported as **did not run** and does not block.
   Nothing was measured, so there is no finding to act on, and CI runs the same
   check on the pull request.

Cost: one `uv run` spawn per gate for the probe, about 2.7s once the
environment is synced. The probe syncs a `.venv` into the worktree as a side
effect — the old code did the same the moment it called `uv run`, so that is
not new.

## Verification

**This PR was pushed from a worktree with the gates enabled** — no
`--no-verify`:

```
CHANGELOG polish + mkdocs --strict + mypy................................Passed
```

A pass is not proof on its own, since the skip path also lets a push through,
so I checked which branch ran. Invoking the hook by hand over the same range in
the worktree prints `pre-push: running mkdocs build --strict ...` and exits 0 —
the gate ran. Both `uv run` probes fail there (`program not found`) and the
main checkout's `.venv` is what served it, i.e. branch 2. mypy correctly did
not run: this branch touches no `factrix/**/*.py`.

`tests/test_pre_push_gate_runner.py` drives the real hook under a stub `uv` in
a throwaway git repo and separates the three outcomes:

| case | expected | on the unfixed hook |
|---|---|---|
| nothing can start the tool | exit 0, "did not run", no "blocked" | **fails** — exit 1, "mkdocs build --strict failed" |
| runner starts it, gate fails | exit 1, "blocked" | passes (false-positive guard) |
| runner starts it, gate passes | exit 0, silent | passes (false-positive guard) |

The first case is parametrised over both gates and is the regression test; the
other two exist so the skip path cannot be widened into swallowing a real
failure. Suite `3855 passed / 45 skipped`, ruff and format clean.

`docs/development/contributing.md` gains the worktree behaviour and a note that
`--no-verify` is for skipping gates that *can* run.

## Not done

The hook file is CRLF in the working tree (no `.gitattributes` pins
`scripts/hooks/*` to `eol=lf`). It works under Git Bash today and is a separate
question from this issue, so it is left alone rather than folded in.

### DIFF
--- a/docs/development/contributing.md
+++ b/docs/development/contributing.md
@@ -106,6 +106,16 @@ to each clone.
 | `commit-msg` | Commit-message and contributor-trailer policy |
 | `pre-push` | Strict docs build for docs-relevant paths, mypy for package paths, and release-note checks where applicable |
 
+The `pre-push` gates run through `uv run`, which resolves the environment of
+the directory the push is issued from. A **git worktree** has no environment of
+its own, so the gates fall back to the main checkout's `.venv`; where neither
+can start a tool the hook says the gate *did not run* and lets the push
+through, rather than reporting it as a gate failure. Nothing was measured in
+that case, and CI still runs the same check on the pull request — but if you
+want the local feedback back, `uv sync --extra docs` inside the worktree gives
+it one of its own. Reach for `git push --no-verify` only when you mean to skip
+gates that *can* run; it drops all of them, including the release-note check.
+
 Run the lint stage directly with:
 
 ```bash
--- a/scripts/hooks/pre-push
+++ b/scripts/hooks/pre-push
@@ -45,6 +45,49 @@ MYPY_PATH_RE='^factrix/.*\.py$'
 needs_mkdocs=0
 needs_mypy=0
 
+# Resolve a command that can actually START a gate tool, printing its words
+# separated by ``|``; return 1 when no environment here can.
+#
+# WHY resolve before running: ``uv run`` resolves the environment of the
+# directory it is invoked from, and a git worktree has none - it exits
+# ``Failed to spawn: mkdocs / program not found`` without mkdocs ever
+# starting. Reporting that as "mkdocs build --strict failed" sends the reader
+# hunting a docs defect that does not exist, and the way out of it is
+# ``--no-verify``, which drops every other gate along with the one that could
+# not run (#1048). The probe costs one ``uv run`` spawn per gate (about 2.7s
+# once the environment is synced); that is the price of the two outcomes
+# staying distinguishable.
+_gate_command() {
+    tool=$1
+    if command -v uv >/dev/null 2>&1 &&
+        uv run --quiet -- "$tool" --version >/dev/null 2>&1; then
+        printf 'uv|run|--quiet|--|%s' "$tool"
+        return 0
+    fi
+    # A worktree shares the main checkout's ``.git``, so the environment the
+    # gate would have used over there is the one to fall back to.
+    common=$(git rev-parse --path-format=absolute --git-common-dir 2>/dev/null || true)
+    root=${common%/.git}
+    for py in "$root/.venv/bin/python" "$root/.venv/Scripts/python.exe"; do
+        if [ -x "$py" ] && "$py" -c "import $tool" >/dev/null 2>&1; then
+            printf '%s|-m|%s' "$py" "$tool"
+            return 0
+        fi
+    done
+    return 1
+}
+
+# Say that the gate DID NOT RUN, which is not the statement that the gate
+# failed, and do not block on it: nothing was measured, so there is no finding
+# to act on, and CI runs the same check on the pull request.
+_gate_unavailable() {
+    printf '\n' >&2
+    printf 'pre-push: %s did not run - no environment here can start it.\n' "$1" >&2
+    printf 'Nothing was measured, so this is not a gate failure. CI still checks it\n' >&2
+    printf 'on the pull request.\n' >&2
+    printf 'From a git worktree: uv sync --extra docs, or push from the main checkout.\n\n' >&2
+}
+
 # Emit ``<local_ref> <local_sha> <remote_ref> <remote_sha>`` lines in
 # Git's pre-push stdin format. Only the two SHA fields are read below,
 # so the ref fields are placeholders when synthesised.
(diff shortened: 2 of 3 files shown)


awwesomeman:
## Review — non-blocking is right, the probe is the right trade. Merging.

Drove the real hook through all three branches from my own worktree rather than
reading the shell.

### Non-blocking for branch 3: yes, and the safety property survives

Your reasoning is right, but the decisive evidence is one you did not state:
**a gate that runs and fails still blocks on the new path too.** I broke
`mkdocs.yml`, forced the fallback with a failing `uv` stub, and pushed:

```
pre-push blocked: mkdocs build --strict failed.
Re-run for the full error: D:/Project/factrix/.venv/Scripts/python.exe -m mkdocs build --strict
EXIT=1
```

So the change does not weaken the gate — it separates "did not run" from
"failed", and only the first is non-blocking. Blocking on branch 3 would make a
tooling problem indistinguishable from a finding, with no finding to act on, and
the only escape is `--no-verify`, which drops the release-note gate too. That
trade is clearly the right way round.

Worth noting the remediation line now names the command that *actually ran*
rather than a generic `uv run` — the repo's own rule applied to the hint, not
just the verdict.

### All three branches, end to end

```
branch 1 (uv can start the tool)          both gates ran, EXIT=0
branch 2 (uv fails, main .venv found)     both gates ran, EXIT=0
branch 3 (neither)                        "did not run" message, EXIT=0
branch 2 + a real mkdocs failure          blocked, EXIT=1
```

Confirmed the probe and the runner agree, which is where this defect class would
have reappeared: the probe asserts `import mkdocs` while the runner uses
`python -m mkdocs`, and those can diverge when a package has no `__main__`. Both
tools answer `--version` under `-m` here, so they do not.
`--git-common-dir` resolves to `D:/Project/factrix/.git` from a worktree and
`${common%/.git}` gives the right root.

One environment note, not a defect: on this host `VIRTUAL_ENV` points at the
main checkout's `.venv`, so **branch 1 served my worktree** — `uv run` found
mkdocs without falling back. Which branch serves a worktree therefore depends on
the developer's shell, and branch 2 is not dead code just because it did not
fire for me. Worth knowing before anyone later reads it as unreachable.

### The probe cost: right trade

Classifying `uv`'s stderr after the fact is exactly the brittle string match this
repo has spent seven PRs removing, and preferring the venv directly would lose
the lock sync in the normal case. ~2.7s against a ~25s `mkdocs build --strict`
is about 10% on the gate it protects. I would not spend your cycles making it
cheaper.

### Your requested mutation

Reverting `scripts/hooks/pre-push` to `origin/main`:

```
2 failed, 2 passed
```

The two "did not run" cases fail with `pre-push blocked: mkdocs build --strict
failed`, and the two guards — ran-and-failed blocks, ran-and-passed is silent —
still pass. The guards hold, so the skip path cannot be widened to swallow a real
failure without the suite noticing.

### The CRLF question — I would not file it

Checked rather than judged by eye:

```
blob:         CR=0  LF=205    -> LF-only
working tree: CR=205          (core.autocrlf=true on this Windows host)
```

The repository stores LF; the CRLF is a local checkout artefact and a Linux or
macOS clone never sees it. And your new test executes `bash scripts/hooks/pre-push`
on every CI run, so a genuine line-ending regression in the blob would fail CI —
stronger coverage than a `.gitattributes` line, which only normalises on
checkout. Deliberately not filed, with that as the reason.

### Verification

`3855 passed, 45 skipped`; `ruff check` and `ruff format --check` clean; `mypy
factrix` clean. CI counted: **19 rollup entries, 17 pass / 2 skipping**,
`mergeStateStatus CLEAN`, head `84a3433c`.

Closes #1048.

Issue #1061, opened 2026-09-04
test(dev): make the documented pytest environment sufficient


awwesomeman:
## Problem

The contributor quick path installs `uv sync --extra dev` and then runs the complete `uv run pytest`. The collected docs inventory tests load the MkDocs Material theme, which belongs to the separate `docs` extra. A dev-only environment therefore fails tests with `Unrecognised theme name: material` even when product tests pass.

Relevant configuration/docs: `pyproject.toml:52-67`, `CONTRIBUTING.md:12-14`, `docs/development/contributing.md:28-33`; CI uses `--all-extras`.

## Expected

Every documented command has a dependency environment that can execute the tests it collects, and optional docs checks are clearly separated if intentionally excluded.

## Definition of Done

- [ ] Choose one contract: dev extra includes dependencies required by collected tests, complete test instructions install dev+docs/all extras, or docs tests skip with an explicit reason when docs dependencies are absent.
- [ ] Root and development contributing guides agree.
- [ ] A fresh documented environment can run its advertised pytest command.
- [ ] CI has a guard for the chosen minimal contributor setup.

## Correction (2026-09-07)

Re-measured on `main` at `02d573b5`: a clean `uv sync --frozen --extra dev` + `uv run pytest` is already green (`3800 passed, 50 skipped`) because `tests/test_docs_paths.py` and `tests/test_docs_inventory_cache_warmup.py` carry module-level `importorskip` guards. The `Unrecognised theme name: material` failure is reachable only in a partial docs environment (mkdocs and mkdocstrings present, `mkdocs-material` absent), where the warmup module guard passes but its two `load_config(mkdocs.yml)` tests abort (`2 failed, 19 passed`). PR #1068 guards those two tests on the `mkdocs-material` distribution, aligns both contributing guides, and adds a `--extra dev` CI job as a pin against drift.

PR #1063, opened 2026-09-04
fix(data): validate complete horizon stamp columns (#1050)

### LINKED ISSUE #1050: fix(data): reject mixed or invalid horizon stamps
## Problem

`_read_int_stamp` and the standalone metric path read only the first value of `_forward_periods` / `_overlap_periods`. A concatenated panel carrying multiple horizons is accepted and the first row silently becomes the horizon for the whole study.

Reproduction: combine rows stamped with forward/overlap periods 1 and 5. `evaluate(..., strict=False)` reports `forward_periods=1` and `overlap_periods=1` instead of rejecting the mixed hypothesis. Row order therefore changes HAC bandwidth, stride, sample floors, result identity, and interpretation.

Relevant code: `factrix/_data_input.py:105-118`, `factrix/metrics/_base.py:333-351`.

## Expected

Every present stamp column is a non-null positive integer constant over the complete input. Forward and overlap stamps are validated together at every public entry point.

## Definition of Done

- [ ] Mixed stamp values raise `UserInputError` before metric dispatch.
- [ ] Null, bool, non-integer, zero, and negative stamps are rejected consistently.
- [ ] `evaluate`, slicing entry points, and standalone metric calls share one validator.
- [ ] Tests prove row order cannot change the resolved horizon.
- [ ] Valid single-horizon panels retain current behavior and metadata.

awwesomeman:
## Summary
- validate both reserved horizon stamp columns over the complete panel
- reject mixed, null, non-integer, zero, and negative stamps before dispatch or slicing
- share the same contract across evaluate, slicing, sample requirements, and standalone metrics

## Test plan
- [x] `python -m pytest -q -p no:faulthandler tests/test_horizon_stamp_validation.py` (21 passed)
- [x] targeted evaluate, standalone metric, sample requirements, and slicing suite (211 passed)
- [x] `ruff check factrix tests/test_horizon_stamp_validation.py`
- [x] `mypy factrix`
- [x] `pytest --doctest-modules factrix` (96 passed, 1 skipped)
- [ ] full pytest: 3816 passed / 50 skipped; 2 docs-inventory failures because this local environment cannot install `mkdocs-material` (#1061). CI installs all extras.

Closes #1050

### DIFF
--- a/docs/api/data-schema.md
+++ b/docs/api/data-schema.md
@@ -130,6 +130,14 @@ metadata does not. A self-attached `forward_return` panel carries neither
 stamp and declares `forward_periods=` (and, on a coarser grid,
 `overlap_periods=`) on `evaluate`.
 
+Every present stamp must remain a non-null positive integer with one value
+over the complete panel. Public entry points reject mixed or invalid stamp
+columns before dispatch (and before `by_slice` partitions the frame), because
+otherwise row or slice order could select a different hypothesis and inference
+configuration. Evaluate different return horizons as separate panels with
+[`evaluate_horizons`](multi-horizon.md); do not concatenate their stamped rows
+into one panel.
+
 ---
 
 ## Common errors
--- a/factrix/__init__.py
+++ b/factrix/__init__.py
@@ -82,8 +82,7 @@ from factrix._data_input import (
     _STAMP_COLUMNS,
     DataInput,
     _coerce_data,
-    _read_forward_periods_stamp,
-    _resolve_forward_periods,
+    _resolve_horizons,
     _resolve_overlap_periods,
     _validate_overlap_periods,
 )
@@ -277,8 +276,7 @@ def evaluate(
     # projection, or to_frame. A self-attached forward_return panel (no stamp)
     # must declare its horizon once via forward_periods= (path B); its
     # overlap defaults to the horizon unless declared via overlap_periods=.
-    fp = _resolve_forward_periods(data, forward_periods)
-    op = _resolve_overlap_periods(data, overlap_periods, horizon=fp)
+    fp, op = _resolve_horizons(data, forward_periods, overlap_periods)
     data = data.drop([c for c in _STAMP_COLUMNS if c in data.columns])
 
     from factrix.metrics._base import MetricBase
@@ -747,7 +745,7 @@ def sample_requirements(
         op = _resolve_overlap_periods(
             data,
             overlap_periods,
-            horizon=_read_forward_periods_stamp(data),
+            horizon=None,
             func_name="sample_requirements",
         )
     elif overlap_periods is not None:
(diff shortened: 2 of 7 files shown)


awwesomeman:
## Review — verified independently, follow-ups collected in 8256f61b

**Guard check (CLAUDE.md "a test that cannot fail proves nothing").** Ran
`tests/test_horizon_stamp_validation.py` against unfixed `511602d8`:
**20 failed, 1 passed**. The one pass is
`test_valid_constant_stamps_preserve_existing_results`, which is the pinning
test — it is now labelled as a guard rather than regression proof.

**Premise checks.** Three things the design could have rested on, measured
rather than reasoned about:

- `evaluate` calls `_resolve_horizons` **after** `_coerce_data`, so the new
  `isinstance(data, pl.DataFrame)` short circuit cannot silently skip a
  coerced frame.
- `by_slice` and `sample_requirements(data=)` are polars-only by contract
  (`_slice_by` raises `TypeError`; the annotation is `pl.DataFrame | None`), so
  the same short circuit opens no "non-polars input skips validation" hole.
- The change from duck-typing to `isinstance` in `_stamped_overlap_periods`
  looked like it could silently drop a pandas panel's stamp. It cannot:
  `quantile_spread(pandas_df)` already fails deeper on `main`
  (`AttributeError: 'DatetimeArray' object has no attribute 'sort'`).
  Not a regression.

Coverage: the other four `_resolve_overlap_periods` call sites
(`slicing/inference.py` x2, `slicing/period_inference.py` x2) pick up the
two-column validation as well. All five DoD boxes on #1050 are met.

**Findings, all fixed in this PR rather than filed:**

1. `_resolve_forward_periods` had no caller left after `_resolve_horizons`
   took over — dead code this PR introduced. `_read_forward_periods_stamp` was
   test-only and defaulted `func_name` to its own private name, which would
   have printed into a user-facing message. Both dropped;
   `_read_overlap_periods_stamp` now requires `func_name`.
2. The mixed-stamp rejection rendered
   `invalid _forward_periods={'n_unique': 2, 'values_preview': [1, 5]}` — a
   diagnostics dict under the column's own key, so the message did not name the
   value the screen read. Now prints `[1, 5]`, with `[...] (5 of N)` past the
   preview cap.
3. Both stamp rejections linked to `api/evaluate#forward_periods-and-overlap_periods`
   while the complete-panel contract this PR wrote lives on `api/data-schema`.
   Retargeted, and "rebuild or de-duplicate the panel" replaced with the remedy
   the docs actually give (separate panels via `evaluate_horizons`).
4. Pinning test labelled; new assertions on the rendered value and docs target,
   plus a preview-cap test. All six fail against `c10b6d6c`.

**Local verification of the follow-up commit:** full suite
`3878 passed, 45 skipped` (no failures — the two docs-inventory failures in the
original test plan did not reproduce here), `pytest --doctest-modules factrix`
`96 passed, 1 skipped`, `ruff check`/`ruff format --check` clean,
`mypy factrix` clean over 83 files.

PR #1064, opened 2026-09-04
fix(event): preserve complete price paths (#1051)

### LINKED ISSUE #1051: fix(event): keep full price paths after forward-return preprocessing
## Problem

`compute_forward_return` drops rows whose forward return cannot be formed, including the final `h+1` price rows. Event path consumers then build their price grid from that already-truncated panel, so valid post-event prices disappear even though the raw data contains them.

Reproduction: on a 100-period one-asset panel with an event at index 70 and `forward_periods=5`, `compute_event_returns(raw, offsets=[24])` returns one observation (`signed_return≈0.14035`); the same call on `compute_forward_return(raw, 5)` returns zero rows. The canonical `evaluate` route requires the latter panel, so users cannot preserve the path through the normal workflow.

Relevant code: `factrix/preprocess/returns.py:521-543`, `factrix/metrics/_primitives/_event_returns.py:104-151`; the same ownership question applies to MFE/MAE path consumers.

This issue owns the data/API boundary for full price paths. The concrete solution and acceptance contract are recorded in the first comment.

awwesomeman:
## Summary

- add a run-level price_data panel so event offsets and MFE/MAE windows retain raw tail and between-evaluation-date prices
- keep event eligibility and forward-return samples owned by data, preventing price-only rows from entering return metrics or sample floors
- report eligible, computed, censored, and reason counts for event offsets and excursion paths
- document the three-grid ownership contract and the pre-1.0 MFE/MAE row-retention migration

## Verification

- pre-commit Ruff hooks passed
- mypy factrix passed (83 source files)
- local full pytest excluding the unavailable mkdocs-material inventory-only file: 3811 passed, 50 skipped
- local doctests: 96 passed, 1 skipped
- focused price-path regressions: 12 passed
- offline sdist and wheel build passed
- GitHub Actions: lint/mypy, Python 3.12-3.14 on Linux/macOS/Windows, dependency floors, doctests, wheel smoke, strict docs build, and link check all passed

Closes #1051

### DIFF
--- a/docs/api/data-schema.md
+++ b/docs/api/data-schema.md
@@ -110,6 +110,13 @@ projected per factor before a metric's keyword arguments are known, so a
 `inspect_data` reports `quantile_spread_vw` as unusable, with a blocker naming
 the column, on a panel that has no `market_cap`.
 
+For event paths, a `price` column on the evaluation panel can only cover that
+panel's retained rows. Pass the original full panel separately as
+[`evaluate(..., price_data=raw)`](evaluate.md#full-price-data-for-event-paths)
+when preprocessing removed the forward-return tail or selected a coarser
+evaluation grid. `data` owns events and return samples; `price_data` owns only
+the complete price grid.
+
 ## Reserved columns
 
 [`compute_forward_return`](preprocess.md) stamps two constant `Int32`
@@ -158,10 +165,10 @@ Full error taxonomy and recovery patterns: [Errors](errors.md).
 The canonical pipeline from raw price/event data to evaluate-ready panel:
 
 ```
-raw price panel  ──compute_forward_return(h)──▶  (date, asset_id, factor, forward_return)
-                                                           │
-                                                           ▼
-                                                        evaluate / by_slice / ...
+raw panel ──compute_forward_return(h)──▶ evaluation + forward-return sample
+    └──────────────── price_data ────────────────┐
+                                                 ▼
+                                           evaluate(...)
 ```
 
 Pre-attachment helpers live in [`factrix.preprocess`](preprocess.md); synthetic panels in [`factrix.datasets`](datasets.md). Wide-format multi-factor inputs are handled by passing the column names through `factor_cols=` on a single `evaluate` call rather than by reshaping the panel.
--- a/docs/api/evaluate.md
+++ b/docs/api/evaluate.md
@@ -17,6 +17,50 @@ title: factrix.evaluate
 
 <hr>
 
+## Full price data for event paths
+
+`compute_forward_return` returns the evaluation panel: it removes rows whose
+forward return cannot be formed and, with `dates=`, keeps only the chosen
+evaluation dates. Those rows are the correct sample for return metrics, but
+they are not a complete price path. A later event offset or MFE/MAE window can
+need prices from the removed tail or from dates between evaluations.
+
+Keep the three grids separate by passing the original panel as `price_data`:
+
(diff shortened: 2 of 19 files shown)


awwesomeman:
## Review fixes pushed (`1e2287a9`)

Three required findings from review, plus three small ones. Follow-ups #1076 (`by_slice` has no `price_data` passthrough) and #1077 (`per_offset` counts published on a withdrawn curve) were filed rather than fixed here — both are outside this PR's boundary.

**1. `mfe_mae` empty-frame reason.** The branch reported `no_price_data`, but since this PR the producer keeps one `missing_price_column` censored row per event, so an empty frame means the panel carried no event at all — a condition the count-zero branch below already distinguishes. It now reports `no_events`. `tests/test_short_circuit_contract_consistency.py` was pinning the old label and is updated. The new guard asserts both reasons as one composite token rather than checking either alone.

**2. Duplicate-key message for `price_data`.** `_coerce_price_data` went through the evaluation panel's gate, so a duplicated price row was explained as a fabricated forward return, which the price panel never feeds. `_normalize_panel` now takes a panel role and selects the explanation; the price panel's says one period holding two prices leaves the excursion walk unable to say which price the event entered at.

**3. Docs.** The "additive API, existing calls keep their prior behavior" claim was false: `factrix/__init__.py:512` makes `evaluate_horizons` pass `price_data=raw` unconditionally. `docs/api/evaluate.md` now states the behaviour change and gives a migration path per caller kind. The period-unit rule is stated on all three pages: `offsets=` and `window=` count periods of whichever grid supplies the walk.

Measured for the docs (evaluation grid keeping every tenth price period, `window=3`):

| Grid the walk read | Median MFE |
|---|---|
| `data` evaluation grid | 0.1272 |
| `price_data` price grid | 0.0120 |

**Small ones.** The `compute_event_returns` re-export now says why it stays out of `__all__` — `mfe_mae` re-exports `compute_mfe_mae` the same way, and the docs page documents the primitive in prose rather than rendering its API twice, which `tests/test_metric_api_members_match_all.py` enforces. The `non_finite_excursion` fallback is reachable only through a hand-assembled frame; a test pins it, **labelled as a pinning test** because it passes on the unchanged codebase.

## Against-unfixed evidence

The three reason-code guards were written first and run on the unfixed tree:

```
FAILED tests/test_mfe_mae.py::TestMfeMae::test_short_circuit_when_empty - AssertionError: assert 'no_price_data' == 'no_events'
FAILED tests/test_mfe_mae.py::TestMfeMae::test_zero_events_and_missing_prices_are_distinct_reasons
FAILED tests/test_short_circuit_contract_consistency.py::test_descriptive_short_circuits_share_shape_reason_and_exception[mfe_mae]
3 failed, 33 passed
```

The duplicate-message guard likewise, with the literal it now forbids confirmed present in the old text:

```
FAILED tests/test_event_price_data.py::test_duplicate_price_rows_are_explained_as_a_price_grid_defect
E   assert 'price_data' in "compute_event_returns(): invalid (date, asset_id)='2 duplicated row(s) of 101'
    Expected: one row per (date, asset_id). The forward return shifts by row position ..."
```

## Verification

Run from the branch with the repository interpreter:

- `ruff check` / `ruff format --check`: clean, 251 files
- `pytest`: **3895 passed, 45 skipped, 0 failed**
- `pytest --doctest-modules factrix`: 96 passed, 1 skipped
- `mkdocs build --strict`: built in 26.27s

Issue #1053, opened 2026-09-04
fix(tradability): compute turnover across changing universes


awwesomeman:
## Problem

`notional_turnover` left-joins previous membership onto current rows and computes `1 - n_kept / n_current`. Securities held previously but absent now never enter the paired frame, so liquidations and survivor reweighting are omitted.

Reproduction: shrink a four-asset universe to two assets with unchanged surviving top/bottom names. An equal-weight leg must sell the departed holding and resize the survivor, giving one-way turnover 0.5 per leg; the current implementation reports 0. Growing universes can conversely misprice new buys under the replacement-cost convention.

Relevant code: `factrix/metrics/tradability.py:603-640`.

## Expected

Turnover is derived from both prior and current holdings on their union, with a documented one-way/two-way convention.

## Definition of Done

- [ ] Prior-only, current-only, and surviving holdings all enter the calculation.
- [ ] Equal-weight leg turnover matches `0.5 * sum(abs(w_t - w_t_minus_1))` or an explicitly equivalent membership definition.
- [ ] Tests cover shrinking, growing, fully rotating, and unchanged universes.
- [ ] Top, bottom, and long-short aggregate metadata remain internally consistent.
- [ ] Cost utilities consume the same turnover convention documented here.

## Correction (2026-09-07)

Measured while fixing (PR #1066): for an equal-weight leg with `j` prior holdings, `k` current holdings and `m` survivors, one-way turnover over the union is `0.5 * sum(abs(w_t - w_t_minus_1)) = 1 - m / max(j, k)`. For `j <= k` (growing or unchanged legs) this collapses to the old `(k - m) / k`, so growing universes were already priced correctly under this convention; only shrinking legs (`j > k`) were wrong, e.g. the four-to-two reproduction reports `0` instead of `0.5`. The "growing universes can conversely misprice new buys" sentence above does not hold.

PR #1070, opened 2026-09-07
fix(tradability): validate cost algebra inputs and unavailable results

### LINKED ISSUE #1054: fix(tradability): validate cost algebra inputs and unavailable results
## Problem

`breakeven_cost` and `net_spread` extract floats without validating their economic domain or propagating unavailable upstream results.

Observed behavior includes `breakeven_cost(0.001, turnover=-0.2, holding_periods=1) -> inf`, while `net_spread(..., turnover=-0.2, estimated_cost_bps=30)` increases alpha. Turnover above 1, negative costs, non-finite values, and `MetricResult(value=NaN)` are also accepted. Zero-turnover break-even always returns positive infinity regardless of gross spread sign.

Relevant code: `factrix/metrics/tradability.py:699-777`, `:914-932`, `:1090-1097`.

## Expected

Cost algebra rejects impossible scalar inputs and preserves the reason/warnings of unavailable producer results.

## Definition of Done

- [ ] Require finite `0 <= turnover <= 1` under the documented turnover definition.
- [ ] Require finite non-negative `estimated_cost_bps` and finite gross spread.
- [ ] Invalid bare scalars raise `UserInputError` consistently at construction/call time.
- [ ] Unavailable `MetricResult` inputs produce a structured unavailable result rather than an applicable-looking NaN.
- [ ] Zero-turnover break-even behavior is defined separately for positive, zero, and negative gross spread.
- [ ] Boundary and property tests cover the full documented domain.

awwesomeman:
> **Stacks on #1066** (`fix/1053-turnover-universe-change`) — both touch `factrix/metrics/tradability.py`, and the turnover domain this PR enforces is the convention #1053 documents. Base is that branch; merge #1066 first and this retargets to `main` cleanly.

## Summary

`breakeven_cost` and `net_spread` extracted floats from their inputs without checking that the numbers described a portfolio that can exist, and turned an unavailable producer result into an applicable-looking NaN.

`_unpack_cost_inputs` now runs three steps in a fixed order, and the order is part of the contract:

1. **Unavailable inputs short-circuit.** A `MetricResult` whose producer declined to publish a number — it carries `METRIC_UNAVAILABLE`, or simply a non-finite `value` — is re-emitted as this metric's own *descriptive* short circuit via `_short_circuit_output`, with `reason` `no_gross_spread` / `no_turnover` (the "missing input" half of that helper's reason vocabulary), the producer's `reason` under `upstream_reason`, its name under `upstream_metric`, and its advisory codes carried along. `gross_spread` is inspected first, so when both are unavailable the spread's reason is the one reported. The non-finite-value check is not redundant with the code: a hand-built `MetricResult(value=NaN)` carries no codes at all.
2. **Domain validation**, raising `UserInputError` through the existing structured-error pattern (`func_name` / `field` / `value` / `expected` / `docs_path`): finite `gross_spread`, finite `turnover` in `[0, 1]` under the #1053 one-way convention, finite non-negative `estimated_cost_bps`. The same bounds apply to a bare scalar and to an *available* `MetricResult`'s `value` — only unavailability takes the structured path, which keeps the two branches from overlapping.
3. **Pairing check** — the existing `n_groups` cross-check, unchanged.

**Zero turnover is no longer one answer for three questions.** A book that trades nothing pays nothing, so the limit is read off the sign of the numerator rather than the ratio:

| `gross_spread` | Result |
|---|---|
| `> 0` | `+inf` — no finite one-way cost takes the alpha to zero |
| `< 0` | `-inf` — no cost `>= 0` makes a book that already loses break even |
| `== 0` | short circuit, `reason="no_unique_breakeven_cost"` — `net` is zero at *every* cost, so no single cost is the boundary |

Returning `+inf` for all three said a losing book can bear an unlimited cost.

Docs: `breakeven_cost` / `net_spread` docstrings (`Raises`, the propagation note, the three zero-turnover cases), a "The cost algebra's domain" section in `docs/api/metrics/tradability.md`, and the short-circuit metadata keys in `docs/reference/stat-keys-by-metric.md`.

## Verification

**Against the unfixed state** (new tests on the #1053 branch's `tradability.py`):

```
$ python -m pytest -q tests/test_tradability.py -k CostAlgebraDomain
...
================ 17 failed, 26 passed, 63 deselected in 1.36s =================
```

Failing rows covered every clause of the DoD — `turnover` at `-0.2` / `1.5` / `nan` / `inf`; non-finite `gross_spread` at `nan` / `inf` / `-inf`; `estimated_cost_bps` at `-5.0` / `nan` / `inf`; an out-of-domain *available* `MetricResult`; the two documented regressions; the three unavailable-propagation cases; and the two new zero-turnover branches:

```
_ TestCostAlgebraDomain.test_unavailable_turnover_propagates_through_both_helpers _
    assert out.metadata["reason"] == "no_turnover"
E   KeyError: 'reason'

_ TestCostAlgebraDomain.test_zero_turnover_with_a_negative_spread_is_minus_infinity _
    assert out.value == float("-inf")
E   AssertionError: assert inf == -inf

_ TestCostAlgebraDomain.test_zero_turnover_with_a_zero_spread_has_no_unique_breakeven _
    assert math.isnan(out.value)
E   assert False
E    +  where False = isnan(inf)
```

The 26 that passed unfixed are the boundary and property rows — the unit-interval endpoints `0.0` / `1.0`, the 12-cell `sign(breakeven) == sign(gross_spread)` sweep over `spread` x `turnover`, the 9-cell `net <= gross` / `cost_drag >= 0` sweep over `turnover` x `cost`, and the descriptive-short-circuit shape. They pin behaviour that was already correct and are labelled as such; the guards are the 17.

After the fix, all 43 pass.

**Gates** (main checkout's interpreter, run from the worktree; the pre-push hook cannot resolve `uv run` from a worktree, so the push used `--no-verify`):

- `ruff check .` — All checks passed
- `ruff format --check .` — 250 files already formatted
- `mypy factrix` — Success: no issues found in 83 source files
- `pytest -q -p no:faulthandler` — **3938 passed, 45 skipped** in 271.32s
- `pytest --doctest-modules factrix -q` — 96 passed, 1 skipped
- `mkdocs build --strict` — Documentation built in 26.75 seconds

Closes #1054

### DIFF
--- a/docs/api/metrics/tradability.md
+++ b/docs/api/metrics/tradability.md
@@ -121,6 +121,49 @@ not two approximations of each other.
     sizes at $t$ — they equal the turnover denominator only while the legs do
     not shrink.
 
+## The cost algebra's domain
+
+`breakeven_cost` and `net_spread` solve one portfolio's arithmetic, so they
+police the economic domain of what they are handed rather than pushing any
+float through the formula.
+
+| Input | Domain | Why |
+|---|---|---|
+| `gross_spread` | finite | A non-finite spread has no reading as a per-period return. |
+| `turnover` | finite, $0 \le \tau \le 1$ | The one-way per-leg replaced fraction above. `rank_turnover` lives in $[0, 2]$ and does not belong here. |
+| `estimated_cost_bps` | finite, $\ge 0$ | A one-way per-trade cost. A negative cost would make trading a source of return. |
+| `holding_periods` | integer $\ge 1$ | A rebalance interval in underlying return periods. |
+
+A violation raises `UserInputError`, with the same bounds applied to a bare
+scalar and to an *available* `MetricResult`'s `value`. Before this was
+enforced, `breakeven_cost(0.001, turnover=-0.2, holding_periods=1)` returned
+$+\infty$ and `net_spread(..., turnover=-0.2, estimated_cost_bps=30)`
+*increased* the alpha it was meant to charge.
+
+!!! note "Unavailable inputs propagate; they are not priced"
+    A `MetricResult` whose producer short-circuited — it carries
+    `METRIC_UNAVAILABLE`, or simply a non-finite `value` — comes back as the
+    consumer's own short circuit: `reason` is `no_gross_spread` or
+    `no_turnover`, the producer's `reason` travels under `upstream_reason`,
+    and its advisory codes are carried along. `gross_spread` is inspected
+    first, so when both are unavailable the spread's reason is the one
+    reported. Running the algebra on a NaN instead yields a NaN breakeven that
+    reads exactly like a computed one.
+
+### Zero turnover — three different questions
+
+A book that trades nothing pays nothing, so `breakeven_cost` reads the limit
+off the sign of the numerator rather than evaluating the ratio.
+
+| `gross_spread` | Result | Reading |
+|---|---|---|
+| $> 0$ | $+\infty$ | The alpha is free to keep; no finite one-way cost takes it to zero. |
+| $< 0$ | $-\infty$ | The book already loses before costs; no cost $\ge 0$ makes it break even. |
+| $= 0$ | short circuit, `reason="no_unique_breakeven_cost"` | `net` is zero at *every* cost, so no single cost is the boundary. |
+
+Returning $+\infty$ for all three — as this function did before — says a
+losing book can bear an unlimited cost.
+
 ## Four period counts, four questions
 
 The tradability surface touches all four of the period counts factrix keeps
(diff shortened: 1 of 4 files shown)


awwesomeman:
CI note: `test.yml` is gated on `pull_request: branches: [main]`, so with this PR based on `fix/1053-turnover-universe-change` only `build` (strict docs) and `pr` (title lint) ran — both pass; `build-main` and `scheduled` are skipping. The pytest / lint / doctest matrix will fire once #1066 merges and this retargets to `main`. Locally on this exact branch: 3938 passed, 45 skipped (full suite), 96 passed / 1 skipped (doctests), ruff + mypy clean.

awwesomeman:
Superseded by #1080. GitHub auto-closed this PR when its base branch was deleted on merging the parent; the same branch is now rebased onto `main` and open as #1080, carrying only its own commit.

Issue #632, opened 2026-06-20
fix(inspect): inspect_data pre-flight verdict accuracy for event metrics


awwesomeman:
## Problem

Two related gaps in \`inspect_data\` cause it to report \`usable=True\` for metrics that will fail at evaluate-time:

**1. \`event_ic\` falsely marked usable for discrete signals**

\`make_event_panel\` produces a canonical ternary factor \`{-1, 0, +1}\`. \`inspect_data\` marks \`event_ic\` as \`usable\` because the cell-match (SPARSE × PANEL) passes and the event-count floor is met. However, \`evaluate\` immediately short-circuits with \`not_applicable_discrete_signal\` because \`event_ic\` requires a continuous magnitude to compute a meaningful IC. The pre-flight check does not inspect signal cardinality, so any user of the standard event panel format will always hit this mismatch.

**2. Multi-factor panel: usable verdict is based only on the first factor column**

When \`inspect_data\` is called on a panel with multiple factor columns (e.g. \`factor_0000\`…\`factor_0054\`), the \`DataProperties\` (density, scope, n\_events) are computed from the first column only. The resulting \`usable\` / \`degraded\` / \`unusable\` partition may not apply to the other columns. A \`CROSS_FACTOR_DENSITY_MISMATCH\` warning is emitted for heterogeneous panels, but the usable list itself is not adjusted.

## Expected

- \`inspect_data\` should detect ternary / low-cardinality discrete signals and mark \`event_ic\` as \`unusable\` (or \`degraded\`) with a clear blocker reason.
- For multi-factor panels, the usable verdict should either reflect the per-column heterogeneity more clearly, or the warning should direct the user to call \`inspect_data(data, factor_cols=[col])\` per column.

## Definition of Done

- [ ] \`inspect_data\` detects discrete / low-cardinality signals and blocks \`event_ic\` accordingly
- [ ] Multi-factor verdict gap is documented or resolved — per-column \`inspect_data\` call is clearly guided when densities differ

PR #1073, opened 2026-09-07
feat(inspect): expose per-factor preflight results

### LINKED ISSUE #1056: feat(inspect): expose per-factor preflight results
## Problem

For multi-factor input, `inspect_data` returns one `DataProperties` object and all metric verdicts are driven by the first factor column. Only scope/density mismatches receive aggregate warnings; event count, magnitude/cardinality, usable pairs, and stage-one profiles for later columns remain hidden.

This behavior was documented as a limitation in #632, but it leaves preflight and `evaluate` at different granularities: `evaluate` dispatches each factor independently while inspection can describe only the first.

Relevant code: `factrix/_inspect.py:571-579`, `:620-646`.

## Expected

A heterogeneous panel has an auditable per-factor inspection result while preserving a concise aggregate view.

## Definition of Done

- [ ] Public output exposes scope, density, counts, profiles, warnings, and metric verdicts per factor.
- [ ] Aggregate warnings identify exactly which columns disagree and on which property.
- [ ] Single-factor callers retain a simple compatible path.
- [ ] Tests cover factors differing in scope, density, missingness, cardinality, and event count.
- [ ] Docs show how preflight results map to multi-factor `evaluate` outputs.

Refs #632

awwesomeman:
> Stacks on #1069 (`fix/1055-scope-ignores-missing-cells`) — both touch `factrix/_inspect.py`. Base is that branch; the diff here is only the per-factor work. Merge #1069 first.

## Summary

`inspect_data` returned one `DataProperties` and one verdict list for a multi-factor panel, both computed from the first column. Event count, magnitude / cardinality, usable pairs and the stage-one profiles of every later column were unreachable, so preflight and `evaluate` — which dispatches each factor column independently — described the panel at different granularities (#632 documented this as a limitation rather than fixing it).

Inspection now runs **per column**. `DataInspection.factors` maps each inspected column to a new public `FactorInspection` carrying that column's scope, density, counts, stage-one profiles, warnings and full metric verdicts, with the same `usable` / `degraded` / `unusable` partitions. Nothing in a `FactorInspection` is inherited from a sibling column.

### Design decisions

**Two granularities, not one.** `properties` / `metrics` and the tier partitions stay the concise aggregate and describe the **first** inspected column — they are exactly `factors[<first column>]`. A single-factor caller therefore never touches the mapping (the simple path is byte-for-byte what it was), and a multi-factor caller never has to guess which column an aggregate refers to. The alternative — making `properties` a mapping — would have broken every single-factor caller for no gain, and a screen over dozens of candidates still wants one answer to "what is this panel".

**Warnings stay attributable.** The aggregate `warnings` carry the first column's warnings plus the cross-factor mismatches. A later column's own advisories (low cardinality, frequent events, scope unidentifiable, sample shape) live on `factors[col].warnings` rather than being merged into one stream where the column they belong to would be lost.

**Mismatch warnings name the disagreement.** They now read `Basis 'factor' (individual); disagreeing columns: 'macro' (common)` and point at `DataInspection.factors[<col>]`, instead of listing every column with its value and advising a re-run per column (which the mapping makes unnecessary). This is a message-text breaking change; the assertions in `test_multi_factor_inconsistent_warnings` are updated.

**One implementation of the shared behaviour.** `DataInspection` and `FactorInspection` inherit one `_MetricPartitionView` — one tier partition and one dict renderer — rather than two that can drift apart.

**`n_periods` / `n_assets` stay panel-level** on every `FactorInspection`: they are properties of the data, and making them per-column would change which sample floor a column is judged against, diverging from the panel-level counts `evaluate` gates on. Where a column's own coverage must gate a metric it does so through the stage-one profile, which *is* per column — that is what separates a column whose IC cross-sections survive on 20 periods (blocked at `ic`'s 50-period floor) from its sibling at 118. Documented explicitly in the docstring and the docs page.

Docs: `docs/api/inspect-data.md` gains a "Multi-factor input: per-factor results" section with a table mapping each preflight field onto the `evaluate` output it predicts, plus a runnable example.

### Cost, measured

Per-column verdicts mean per-column work, so it was measured rather than assumed — on a 100-asset × 500-period panel (49,400 rows) with 55 factor columns:

| | 1 column | 55 columns |
|---|---|---|
| before | 0.031 s | 0.245 s |
| after | 0.029 s | 0.937 s |

~17 ms per column for a full per-column verdict; the single-column path is unchanged. The first draft measured 0.32 s for **one** column — the profile and signal-shape scans had been left inside the per-spec comprehension and ran ~40× per column. That is in the diff as a hoist with a comment; without measuring it would have shipped as a 10× regression on the single-factor path.

## Verification

Gates (run manually from the worktree; the pre-push hook cannot resolve tools there, so the push used `--no-verify`):

- `ruff check .` → `All checks passed!`
- `ruff format --check .` → `250 files already formatted`
- `mypy factrix` → `Success: no issues found in 83 source files`
- `pytest -q` → `3903 passed, 44 skipped, 451 warnings in 252.63s`
- `pytest --doctest-modules factrix -q` → `96 passed, 1 skipped`
- `mkdocs build --strict` → `Documentation built in 26.24 seconds`

### Against the unfixed state

Tests were written first. With `git stash push -- factrix` (implementation reverted, tests kept), the new class and the updated mismatch assertions fail 14/14:

```
FAILED tests/test_inspect_data.py::TestCrossFactorConsistency::test_multi_factor_inconsistent_warnings
FAILED tests/test_inspect_data.py::TestPerFactorInspection::test_factors_covers_every_inspected_column_in_order
FAILED tests/test_inspect_data.py::TestPerFactorInspection::test_single_factor_path_stays_simple
FAILED tests/test_inspect_data.py::TestPerFactorInspection::test_aggregate_still_describes_the_first_column
FAILED tests/test_inspect_data.py::TestPerFactorInspection::test_scope_is_reported_per_factor
FAILED tests/test_inspect_data.py::TestPerFactorInspection::test_density_is_reported_per_factor
FAILED tests/test_inspect_data.py::TestPerFactorInspection::test_missingness_is_reported_per_factor
FAILED tests/test_inspect_data.py::TestPerFactorInspection::test_event_count_is_reported_per_factor
FAILED tests/test_inspect_data.py::TestPerFactorInspection::test_cardinality_verdict_is_reported_per_factor
FAILED tests/test_inspect_data.py::TestPerFactorInspection::test_metric_verdicts_differ_per_factor
FAILED tests/test_inspect_data.py::TestPerFactorInspection::test_per_factor_stage_one_profile_gates_its_own_column
FAILED tests/test_inspect_data.py::TestPerFactorInspection::test_aggregate_warning_names_the_disagreeing_columns
FAILED tests/test_inspect_data.py::TestPerFactorInspection::test_to_dict_carries_every_factor
FAILED tests/test_inspect_data.py::TestPerFactorInspection::test_repr_html_lists_each_factor_with_its_own_axes
====================== 14 failed, 80 deselected in 1.18s ======================
```

Coverage per the DoD — one panel carrying four deliberately different columns (individual dense, broadcast dense, zero-encoded sparse, low-cardinality with holes) plus a stage-one-thin column: scope, density, missingness (`n_pairs`), event count (`n_events`), cardinality (the low-cardinality advisory lands on its own column and not on the first), differing metric verdicts (`ic` usable on the individual column and unusable on the common one, `common_beta` the reverse), per-column stage-one gating, the single-factor simple path, `to_dict`, `_repr_html_` and the mismatch messages.

Two tests were strengthened after first drafts passed unfixed: the `_repr_html_` test asserted loose substrings that already appeared in metric names, and the stage-one test first used a construction that gated on scope rather than the profile. Both now assert whole rows / the exact blocker string.

## Follow-up found while writing the docs example (not fixed here)

`factors[col].usable.to_metrics_dict()` fed straight to `evaluate(strict=True)` still raises on a broadcast common column: `common_beta_profile` / `common_beta_r_squared` / `common_beta_sign_consistency` report `n_assets=0` at run time while preflight calls them usable, and `spanning_alpha` needs base factors that no preflight axis models. That is a pre-existing preflight/run-time gap on the common-scope cell — `tests/test_inspect_evaluate_agreement.py` sweeps individual dense panels only — and is out of this PR's boundary. The docs example uses `strict=False`. I will file it as its own issue.

Closes #1056
Refs #632

### DIFF
--- a/docs/api/inspect-data.md
+++ b/docs/api/inspect-data.md
@@ -84,7 +84,76 @@ The tier groups are `MetricApplicabilityGroup` objects: they expose `.names`
 and `.to_metrics_dict()`, and slicing or concatenating them with `+` preserves
 those helpers.
 
+<hr>
+
+## Multi-factor input: per-factor results
+
+`evaluate` dispatches each factor column independently — one
+`EvaluationResult` per column, each routed to its own `(scope, density,
+structure)` cell. Inspection reports at the same granularity, so a preflight
+verdict maps one-to-one onto the `evaluate` output it predicts:
+
+| `inspect_data(data, factor_cols=cols)` | `evaluate(data, ..., factor_cols=cols)` |
+|---|---|
+| `info.factors[col].properties` | the cell `results[col]` was dispatched to |
+| `info.factors[col].usable` / `.degraded` / `.unusable` | which entries of `results[col].metrics` return a value, a degraded value, or a `metric_unavailable` NaN |
+| `info.factors[col].warnings` | the data-level warnings that column raises |
+| `info.factors[col].usable.to_metrics_dict()` | the `metrics=` argument to run that column safely |
+
+```python
+import factrix as fx
+import polars as pl
+from factrix.preprocess import compute_forward_return
+
+raw = fx.datasets.make_cs_panel(n_assets=20, n_dates=120)
+data = compute_forward_return(
+    raw.with_columns(pl.col("factor").mean().over("date").alias("macro")),
+    forward_periods=5,
+)
+info = fx.inspect_data(data, factor_cols=["factor", "macro"])
+
+for col, f in info.factors.items():
+    print(col, f.properties.scope.value, f.properties.density.value, f.usable.names)
+
+# Run each column with the metrics its own verdict cleared. strict=False keeps
+# a metric whose run-time floor binds tighter than the pre-flight one as a NaN
+# placeholder with a reason, rather than raising for the whole batch.
+for col, f in info.factors.items():
+    results = fx.evaluate(
+        data,
+        metrics=f.usable.to_metrics_dict(),
+        factor_cols=[col],
+        forward_periods=5,
+        strict=False,
+    )
+```
+
+Two granularities travel together on purpose. A screen over dozens of
+candidate columns still wants one answer to "what is this panel", so
+`properties`, `metrics` and the tier partitions remain a concise aggregate and
+describe the **first** inspected column — exactly `factors[<first column>]`.
+Single-factor callers therefore never touch the mapping; multi-factor callers
+never have to guess which column an aggregate refers to.
+
+When columns disagree on `FactorScope` or `FactorDensity`, the aggregate
+`warnings` carry a `cross_factor_*_mismatch` naming the basis column, the
+columns that disagree with it, and the value each one carries. A later
+column's own advisories (low cardinality, frequent events, scope
+unidentifiable, sample shape) live on `factors[col].warnings` rather than
+being merged into one stream where the column they belong to would be lost.
+
+One panel-level caveat: `n_periods` and `n_assets` are properties of the data,
+so every `FactorInspection` reports the panel's counts rather than the periods
+that column happens to cover. Where a column's own coverage gates a metric it
+does so through the stage-one profile, which *is* per column — a column whose
+IC cross-sections survive on 20 periods is blocked at `ic`'s 50-period floor
+while its sibling in the same panel is not.
+
 ::: factrix.DataInspection
+
+---
+
+::: factrix.FactorInspection
 
 ---
 
--- a/factrix/__init__.py
+++ b/factrix/__init__.py
@@ -96,6 +96,7 @@ from factrix._errors import (
 from factrix._inspect import (
     DataInspection,
     DataProperties,
+    FactorInspection,
     MetricApplicability,
     MetricApplicabilityGroup,
     _allows_sparse_event_override,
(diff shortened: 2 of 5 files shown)


awwesomeman:
CI note: `.github/workflows/test.yml` triggers only on `pull_request: branches: [main]`, so with this PR based on `fix/1055-scope-ignores-missing-cells` the rollup carries just 4 entries — `build` pass, `pr` pass, `build-main` and `scheduled` skipped. **No pytest, lint, mypy or doctest job ran on this PR**; the rollup being green is a stacking artifact, not test coverage. The full local gates are in the PR body. Retarget to `main` after #1069 merges to get the real matrix.

awwesomeman:
Superseded by #1081. GitHub auto-closed this PR when its base branch was deleted on merging the parent; the same branch is now rebased onto `main` and open as #1081, carrying only its own commit.

PR #1075, opened 2026-09-07
feat(oos_decay): provide a first-class robust split workflow

### LINKED ISSUE #1059: feat(oos_decay): provide a first-class robust split workflow
## Problem

`oos_decay` exposes one fixed row-count split. Documentation recommends sweeping 60/40, 70/30, and 80/20 caller-side and cites purged/CPCV approaches, but the library does not aggregate those runs, enforce a pre-declared split set, prevent overlapping-horizon leakage at boundaries, or record selection provenance.

The current primitive is useful, but a single arbitrary cut point can reverse the gate when a regime change lands near that point. Ad hoc caller loops also make multiplicity and sign-flip policy inconsistent across users.

Relevant code: `factrix/metrics/oos_decay.py:4-8`, `:90-159`, `:197-246`; `docs/api/metrics/oos_decay.md:34-55`.

## Expected

Provide a small, explicit robustness workflow built on the existing primitive rather than silently replacing its semantics.

## Definition of Done

- [ ] Accept or define a pre-declared set of split fractions and aggregate rule.
- [ ] Document whether any sign flip vetoes and how survival ratios are summarized.
- [ ] Account for `forward_periods`/overlap when separating IS and OOS windows, or clearly state why no purge applies.
- [ ] Return every split result plus aggregate status and provenance.
- [ ] Tests cover a structural break near one cut point and confirm order-invariant aggregation.
- [ ] Clarify multiplicity limits: this is robustness validation, not an additional uncorrected model search.

awwesomeman:
> **Stacked on #1065** (`fix/1058-oos-decay-input-contract`). Both touch
> `factrix/metrics/oos_decay.py`; this PR targets that branch, so review the
> diff against it rather than against `main`. Merge #1065 first.

## Summary

`oos_decay` exposes one fixed cut point. Measured on a 100-period series with
a structural break parked at 0.7 (the fixture this PR adds):

```
is_ratio=0.6: survival=1.0000 status=PASS
is_ratio=0.7: survival=0.0000 status=VETOED
is_ratio=0.8: survival=2.6667 status=PASS
```

One arbitrary fraction decides the gate. The docs told callers to sweep the
fraction themselves, which left the aggregate rule, the sign-flip policy, the
overlapping-horizon purge and the provenance to be reinvented per caller.

`oos_decay_splits` is that sweep as one entry point built on the **unchanged**
primitive — no `oos_decay` semantics move in this PR.

### Design decisions

**Declared split set.** `split_fractions` (default `(0.6, 0.7, 0.8)`) must be
non-empty, distinct, and each strictly inside `(0, 1)`. It is canonicalised
ascending, so the order the caller declared them in cannot move the verdict. A
repeated fraction is rejected rather than de-duplicated — it would
double-weight one cut point inside the aggregate without changing the reported
`n_splits`.

**Aggregate rule — median, and it is not a knob.** `status="PASS"` requires all
three of: every declared split assessable; **no** split sign-flipped; the
**median** survival ratio at or above `survival_threshold`.

Direction and magnitude are aggregated differently on purpose. A sign flip says
the factor predicts the wrong way over some contiguous tail, which the
single-split primitive already treats as a hard veto — a majority vote across
splits would launder it. A magnitude below the bar at one cut point is exactly
the arbitrariness the median is there to absorb (50 % breakdown point, and
order-invariant). The median is fixed rather than exposed as an `aggregate=`
knob because choosing an aggregate *after* seeing the per-split ratios is the
search this workflow exists to close off.

When the gate vetoes on a sign flip, `value` still carries the median that ran
— the veto lives in `status` / `n_sign_flips`, not in a corrupted number.

**Unassessable splits withhold the aggregate.** If any declared split produces
no finite ratio, `value` is NaN with `reason="unassessable_splits"` and
`status="VETOED"`. The aggregate is defined over the set that was *declared*;
recomputing it over whichever subset happened to work would report a silently
narrowed sample as a verdict, and "cannot assess" must not read as "passed".
Every split is still returned for diagnosis.

**Purge.** `forward_periods` periods are dropped off the end of each in-sample
window, counted on the distinct-date grid (never calendar time). A value
stamped at period `t` built from a `forward_periods`-period forward return is
realised over `(t, t + forward_periods]`, so without the gap the last in-sample
observations are partly realised inside the out-of-sample window. The gap comes
off the **in-sample** side only: the out-of-sample window is the thing being
validated and is never shortened to protect the window it is validated against.
A cut whose in-sample side the purge empties reports its own
`reason="purged_split_too_short"` rather than borrowing the primitive's
`insufficient_oos_periods`, which means "the series is too short" — a reader
who sees it needs a nearer cut point or a shorter horizon, not more data.

**Provenance.** `metadata["splits"]` carries one record per declared fraction
(ascending) with `split_fraction`, `n_is`, `n_oos`, `purge_periods`,
`survival`, `sign_flipped`, `status`, plus `mean_is` / `mean_oos` / `reason` /
`signal_status` where the primitive emitted them. Beside it: `status`,
`split_fractions`, `n_splits`, `n_assessable`, `n_sign_flips`, `aggregate`,
`sign_flip_policy`, `forward_periods`, `purge_periods`, `survival_threshold`.
The per-split `n_is` is read back off the split the primitive actually cut at,
not the one requested. No `p_value` is emitted anywhere.

**Multiplicity.** Documented on the page as a warning admonition: the split set
is declared up front, every split is reported, the aggregate is fixed, and no
`p_value` exists — so there is nothing to correct and no best split to select.
Re-running with different fraction sets until one passes converts it into an
uncorrected search; the pre-declaration is the only thing preventing that, and
the library cannot enforce it.

**Placement.** A plain function in `factrix/metrics/oos_decay.py`, exported
through `factrix.metrics.__all__` — not `factrix/__init__.py`, which exports
cross-metric workflows (`by_slice`, `compare`) but no metrics-package symbol,
so top-level export would be the inconsistent choice here. It is deliberately
**not** a registered `@metric`: it has no `MetricSpec` and does not route
through `evaluate`, because it is a caller-side robustness read over a metric,
not a new statistic in the cell taxonomy. Since it has no `@metric` constructor
hook, its knob contract lives in one `_validate_oos_decay_splits` chokepoint
called before any data work — the same single-site rule
`test_no_metric_body_calls_a_knob_validator` enforces for registered metrics.

## Verification

All gates run from the worktree with `.venv/Scripts/python.exe` (`PYTHONUTF8=1`).

- `ruff check .` — All checks passed
- `ruff format --check .` — 250 files already formatted
- `mypy factrix` — Success: no issues found in 83 source files
- `pytest -q -p no:faulthandler` — **3920 passed, 49 skipped** in 232s
- `pytest --doctest-modules factrix -q` — 97 passed, 1 skipped
- `mkdocs build --strict` — Documentation built in 24.18 seconds

### The new tests fail against the unfixed state

**1. The symbol does not exist on the base branch**, so the whole file fails to
collect — necessary, but weak evidence on its own:

```
E   ImportError: cannot import name 'oos_decay_splits' from 'factrix.metrics.oos_decay'
=========================== short test summary info ===========================
ERROR tests/test_oos.py
!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!
```

**2. So each behavioural guard was checked against its own defect**, by
disabling the mechanism it claims to guard (`is_end = split_idx` — purge off —
and `return declared` — canonicalisation off) and re-running:

```
E   comparison failed
      Obtained: 0.013820335636722605
      Expected: 1.0 ± 1.0e-06
    tests/test_oos.py:444   (test_purged_periods_cannot_reach_the_statistic)
E   assert (70, 30) == (69, 30)      tests/test_oos.py:455
E   assert False  where False = isnan(0.9282051282051282)   tests/test_oos.py:519

FAILED tests/test_oos.py::TestSplitSweepContract::test_aggregation_is_order_invariant
FAILED tests/test_oos.py::TestSplitSweepPurge::test_purge_shortens_is_only
FAILED tests/test_oos.py::TestSplitSweepPurge::test_purged_periods_cannot_reach_the_statistic
FAILED tests/test_oos.py::TestSplitSweepPurge::test_a_one_period_purge_still_drops_a_period
FAILED tests/test_oos.py::TestSplitSweepUnassessable::test_a_short_circuited_split_withholds_the_aggregate
================= 5 failed, 21 passed, 34 deselected in 0.20s =================
```

### One guard was vacuous and was rewritten

`test_purged_periods_cannot_reach_the_statistic` originally poisoned the purge
gap of the `_BREAK_AT_070` fixture. Measured directly, it proved nothing:

```
purge=0 clean survival = 0.000000
purge=0 dirty survival = 0.000000
differ: False
```

On that fixture the out-of-sample mean is exactly 0, so `survival = 0 / |mean_is|`
is 0 whatever the in-sample tail contains — the test would have passed with the
purge deleted. It now runs on a flat `_PURGE_PROBE` series and asserts **both**
halves: that the poisoned periods *do* move the un-purged single-split answer
(`1.0000 → 0.0138`, PASS → VETOED), and that they leave the purged sweep at
`1.0`. The first assertion is what stops the second from passing vacuously.

### Structural break, measured

```
--- single-split primitive, per fraction (no purge) ---
  is_ratio=0.6: survival=1.0000 status=PASS
  is_ratio=0.7: survival=0.0000 status=VETOED
  is_ratio=0.8: survival=2.6667 status=PASS
  sweep: median=1.0000 status=PASS
```

`test_structural_break_near_one_cut_point_does_not_decide_the_gate` pins exactly
this contrast. `test_any_sign_flip_vetoes_even_when_the_median_survives` pins
the other direction on a near-identical fixture whose 0.7 cut flips sign: the
median is 0.975 — comfortably above the 0.5 threshold — and the gate still
VETOES, so the flip policy is doing the work rather than the magnitude.

### Incidental

`n_sign_flipped` was renamed to `n_sign_flips` before landing: it matched the
FX008 short-circuit-flag grammar (`*_flipped`) and would have been counted as a
stage flag by `test_fx008_docstring_count_matches_the_grammar`. It is a count,
not a stage flag, so renaming the key beat widening that lint's ledger for a
false positive.

Closes #1059

### DIFF
--- a/docs/api/metrics/oos_decay.md
+++ b/docs/api/metrics/oos_decay.md
@@ -7,6 +7,7 @@ title: factrix.metrics.oos_decay
       show_root_members_full_path: true
       members:
         - oos_decay
+        - oos_decay_splits
 
 <hr>
 
@@ -68,22 +69,77 @@ title: factrix.metrics.oos_decay
     decay around 32 %; factrix's default `survival_threshold = 0.5`
     sits inside that window.
 
--   __Sweep the split fraction caller-side__
+-   __Sweep the split fraction with `oos_decay_splits`__
 
     ---
 
-    One call is one `is_ratio`. To read survival across several
-    fractions, call `oos_decay` per fraction and aggregate yourself —
-    the median absorbs a regime change that lands inside one split,
-    where the mean would not.
+    One `oos_decay` call is one `is_ratio`, and a regime change landing
+    near that cut point can reverse the gate.
+    `oos_decay_splits` runs the same primitive over a set of fractions
+    the caller declares up front (default `(0.6, 0.7, 0.8)`), purges the
+    overlapping in-sample tail, and reports one aggregate verdict plus
+    every individual split.
 
 </div>
+
+## Robustness sweep — `oos_decay_splits`
+
+!!! warning "Robustness validation, not another model search"
+    The split set is declared **before** the sweep runs, every split is
+    reported, the aggregate rule is fixed, and no $p$-value is emitted
+    anywhere — so there is nothing to correct for and no best split to
+    select. Re-running with different fraction sets until one passes turns
+    it back into an uncorrected search. The pre-declaration is the only
+    thing preventing that, and the library cannot enforce it for you.
+
+**Aggregate rule.** `status = "PASS"` requires all three of:
+
+1. every declared split assessable (a split that could not be computed
+   withholds the aggregate: `value` is NaN, `reason = "unassessable_splits"`
+   — "cannot assess" must not read as "passed");
+2. **no** split sign-flipped;
+3. the **median** survival ratio at or above `survival_threshold`.
+
+Direction and magnitude are aggregated differently on purpose. A sign flip
+says the factor predicts the wrong way over some contiguous tail, which the
+single-split primitive already treats as a hard veto — a majority vote over
+splits would launder it. A magnitude below the bar at one cut point is
+exactly the arbitrariness the median is there to absorb (50 % breakdown
+point, and order-invariant, so the declaration order cannot move the
+verdict). The median is fixed rather than exposed as a knob: choosing an
+aggregate after seeing the per-split ratios is the search this closes off.
+When the gate vetoes on a sign flip, `value` still carries the median that
+ran — the veto lives in `status` and `n_sign_flips`, not in a corrupted
+number.
+
+**Purge gap.** A value stamped at period $t$ built from a
+`forward_periods`-period forward return is realised over
+$(t,\, t + \texttt{forward\_periods}]$, so the last in-sample observations
+are partly realised inside the out-of-sample window. `oos_decay_splits`
+drops `forward_periods` periods off the end of each in-sample window —
+counted on the panel's distinct-date grid, never calendar time — which is
+the purge of Lopez de Prado (2018). The gap comes off the **in-sample** side
+only: the out-of-sample window is the thing being validated and is never
+shortened to protect the window it is validated against.
+
+```python title="Illustrative"
+from factrix.metrics import oos_decay_splits
+
+out = oos_decay_splits(
+    ic_df, value_col="ic", split_fractions=(0.6, 0.7, 0.8), forward_periods=5,
+)
+print(out.value, out.metadata["status"])           # median ratio, aggregate gate
+for split in out.metadata["splits"]:               # provenance, ascending
+    print(split["split_fraction"], split["n_is"], split["n_oos"],
+          split["survival"], split["sign_flipped"], split["status"])
+```
 
 ## Choosing a function
 
 | Goal                                                                          | Function                |
 |-------------------------------------------------------------------------------|-------------------------|
 | Single-split OOS survival + sign-flip gate on a `(date, value)` series        | `oos_decay` |
+| Purged survival across a pre-declared set of splits, with provenance          | `oos_decay_splits` |
 
 ## Worked example — IC series fed into oos_decay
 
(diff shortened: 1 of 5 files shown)


awwesomeman:
Superseded by #1082. GitHub auto-closed this PR when its base branch was deleted on merging the parent; the same branch is now rebased onto `main` and open as #1082, carrying only its own commit.

PR #1078, opened 2026-09-07
fix(event): benchmark each offset over its own horizon (#1052)

### LINKED ISSUE #1052: fix(event): align post-event excess benchmark with return horizon
## Problem

`event_around_return` subtracts one unconditional single-bar return from every offset. Negative offsets are single-bar returns, but positive offset `k` is a cumulative `k`-period return from `t+1` to `t+1+k`.

On a pure-drift price process, the reported post-event excess at `k` therefore retains roughly `(k-1)` bars of drift and can look like event alpha. The helper docstring also says the baseline is computed per asset then pooled, while the implementation pools all bar observations directly, overweighting longer-history assets on ragged panels.

Relevant code: `factrix/metrics/event_horizon.py:50-85`, `factrix/metrics/event_horizon.py:242-268`, `factrix/metrics/_primitives/_event_returns.py:142-155`.

## Expected

Every excess-return summary uses a benchmark with the same horizon and weighting unit as the return being summarized.

## Definition of Done

- [ ] Positive offset `k` uses a matching `k`-period benchmark or is explicitly reported as raw cumulative return.
- [ ] Non-positive offsets retain a one-bar benchmark.
- [ ] Baseline weighting matches the documented per-asset/panel contract.
- [ ] A deterministic drift-only panel produces approximately zero excess at all offsets.
- [ ] Docs distinguish raw cumulative, abnormal, and leakage quantities.

awwesomeman:
Stacked on #1064 (`fix/1051-event-price-data`), which owns the price-grid work this measures on. Retarget to `main` once #1064 merges.

## Summary

- offset `k > 0` is measured against `(1 + baseline)**k - 1`, the baseline compounded over the periods that offset actually spans; offsets at or below zero keep the single-period baseline
- each `per_offset` entry publishes the `benchmark` subtracted from it, so the arithmetic is reproducible from the metadata
- `baseline_bar_return` is now one mean per asset pooled with equal weight, matching what its docstring already claimed, with `n_assets_in_baseline` reporting how the weighting was formed
- docs distinguish raw cumulative return, abnormal return, and the leakage score, and say which one the curve publishes

## The defect

`event_around_return` subtracted one unconditional single-period return from every offset. That is the right unit for the pre-event offsets, which are single-period returns, but offset `k > 0` is a simple return from `t+1` to `t+1+k`. Subtracting one period of drift from a `k`-period return leaves roughly `k - 1` periods of it in the answer, and on a panel that merely trends the residue reads as event alpha.

Separately, the baseline helper's docstring described a per-asset mean pooled across assets while the implementation pooled every period observation directly, so a long-history name outvoted a short one on a ragged panel.

## Against-unfixed evidence

The three guards were written first and run on the unfixed tree:

```
FAILED tests/test_event_benchmark_horizon.py::test_drift_only_panel_has_no_excess_at_any_offset
E   AssertionError: offset 6 keeps 0.0202413 of pure drift
E   assert 0.02024128384614812 == 0.0 ± 1.0e-12
FAILED tests/test_event_benchmark_horizon.py::test_baseline_weights_assets_equally_on_a_ragged_panel
E   Obtained: 0.005311594202898546
E   Expected: 0.0034999999999999476 ± 3.5e-12
FAILED tests/test_event_benchmark_horizon.py::test_per_offset_reports_the_benchmark_it_subtracted
E   KeyError: 'benchmark'
3 failed, 1 warning
```

The drift-only panel compounds every asset at exactly 0.4% a period, so no event carries information and every offset prices to zero by construction. The residue at offset 6 is `(1.004**6 - 1) - 0.004 = 0.0202`, i.e. the five extra periods of drift the single-period benchmark left behind. All three pass after the fix.

The weighting guard pins two assets with different drifts and very different history lengths: equal-weighted gives 0.0035, the pooled mean gave 0.00531 — sitting near the long asset's 0.006 rather than describing the panel.

## Verification

Run from this branch with the repository interpreter:

- `ruff check` / `ruff format --check`: clean, 252 files
- `pytest`: **3898 passed, 45 skipped, 0 failed**
- `pytest --doctest-modules factrix`: 96 passed, 1 skipped
- `mkdocs build --strict`: built in 24.45s

`.github/workflows/test.yml` triggers only on `pull_request: branches: [main]`, so the pytest and lint matrix does not run while this PR targets a stacked base. Its small green rollup is a stacking artifact; the matrix fires on retarget.

Closes #1052

### DIFF
--- a/docs/api/metrics/event_horizon.md
+++ b/docs/api/metrics/event_horizon.md
@@ -26,6 +26,31 @@ title: factrix.metrics.event_horizon
     the default factrix presentation; downstream consumers should not
     re-cumulate the pre-event leg.
 
+!!! info "Three quantities, one curve"
+    `per_offset[k]` reports an **abnormal** return, never a raw one, and the
+    metric's headline `value` is a third quantity again. Keeping them apart
+    is the difference between reading event alpha and reading the drift the
+    asset had anyway.
+
+    | Quantity | What it is | Where it appears |
+    |---|---|---|
+    | Raw cumulative return | The formula in the table above, sign-adjusted, with nothing subtracted | Not published; recover it as `mean + benchmark` |
+    | Abnormal (excess) return | Raw return minus that offset's `benchmark` | `per_offset[k]["mean"]`, and every dispersion key beside it |
+    | Leakage score | Mean of the absolute abnormal returns over the *negative* offsets | the metric's `value` |
+
+    The `benchmark` is the panel's unconditional return **over the same
+    horizon as the offset it is subtracted from**: `baseline_bar_return` for
+    offsets at or below zero, and `(1 + baseline_bar_return)**k - 1` for
+    offset $k > 0$, which is a $k$-period return. Subtracting one period of
+    drift from a $k$-period return leaves roughly $k - 1$ periods of it in
+    the answer, and on a panel that merely trends that residue reads as event
+    alpha. On a pure-drift panel every offset now prices to zero.
+
+    `baseline_bar_return` is formed as one mean per asset, pooled with equal
+    weight, and `n_assets_in_baseline` reports how many assets entered it.
+    Pooling every period observation instead would let a long-history name
+    outvote a short one on a ragged panel.
+
 !!! warning "Descriptive only — no p-value is produced"
     `event_around_return` runs no hypothesis test: `p_value` is `None`, and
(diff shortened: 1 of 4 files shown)


awwesomeman:
Superseded by #1079. GitHub auto-closed this PR when its base branch was deleted on merging the parent; the same branch is now rebased onto `main` and open as #1079, carrying only its own commit.

PR #1083, opened 2026-09-07
fix(slicing): pass price_data through by_slice

### LINKED ISSUE #1076: fix(slicing): pass price_data through by_slice so event paths keep the full price grid
## Problem

PR #1064 (#1051) added a run-level `price_data` panel so `event_around_return` and `mfe_mae` walk the complete price grid. `by_slice` (`factrix/slicing/dispatcher.py`, around line 32) has no `price_data` passthrough, so on the slicing route event paths are still truncated to the evaluation panel: tail bars after the last evaluation period and bars between evaluation periods are lost exactly as before #1051.

Found during review of #1064; out of that PR's boundary.

## Expected

`by_slice` accepts `price_data` and forwards it to every slice, and each slice's event metrics report the same eligible/computed/censored counts they would report on the unsliced route.

## Definition of Done

- [ ] `by_slice` accepts and forwards `price_data`; the slice panel keeps eligibility ownership (price-only rows never enter return metrics or sample floors).
- [ ] A regression test shows an offset lost on the sliced route without `price_data` and restored with it; run it against the unfixed state first.
- [ ] `docs/api/slicing.md` (or the equivalent page) documents the parameter with the same period-unit note as `docs/api/evaluate.md`.

Refs #1051 #1064

awwesomeman:
## Summary

- add `price_data` to `by_slice` and forward the complete price grid to every per-slice `evaluate` call
- keep sliced `data` authoritative for event eligibility and evaluation sample counts
- add a regression test that restores a tail offset and matches the direct `evaluate` audit
- document price-grid period units in the slicing API and LLM reference

## Verification

- `uv run pre-commit run --all-files --hook-stage pre-commit`
- `uv run mypy factrix`
- `uv run pytest -p no:faulthandler tests/test_by_slice.py tests/test_event_price_data.py -q` (39 passed)
- `uv run pytest -p no:faulthandler tests/test_docs_examples.py tests/test_docs_llms.py -q` (88 passed, 44 skipped)

Closes #1076


awwesomeman:
## Review

The parameter, the forwarding and the docs are right, and CI is green (20 checks: 18 success, 2 skipped; full suite re-run locally on the branch, 4077 passed / 44 skipped / 0 failed). But `price_data` is forwarded **un-sliced**, and two per-slice quantities are read off the whole price panel rather than the slice. Both are silent wrong answers, and both have the same one-line cause.

### Required 1 — the excess benchmark leaks across slices

`event_around_return` computes `baseline_bar_return` from `path_data`, which becomes the whole `price_data` panel when one is supplied. So every slice is measured against the drift of assets that are not in it.

Measured on a 12-asset panel split into two cohorts, one drifting 0.05% a period and one 0.80%, with a single event per asset and no event information at all (`offsets=[6]`):

| Call | Slice | `baseline_bar_return` | `n_assets_in_baseline` | offset-6 excess |
|---|---|---|---|---|
| no `price_data` | slow | 0.000500 | 6 | 0.0 |
| no `price_data` | fast | 0.008000 | 6 | 2.2e-16 |
| `price_data=raw` | slow | 0.004250 | 12 | **-0.022769** |
| `price_data=raw` | fast | 0.004250 | 12 | **+0.023198** |

Passing `price_data` fabricates ±2.3% of event alpha on a panel where the correct answer is zero, because both slices are handed the pooled 0.00425 baseline. This is the defect class #1052 just closed — a benchmark that does not match the sample it is subtracted from — reintroduced one level up.

It also contradicts this PR's own stated contract. The docstring says the side panel "only supplies the complete price grid for event offsets and excursion windows", but it is also supplying the baseline and the assets behind it.

### Required 2 — the raggedness verdict leaks across slices

Same cause. `_warn_ragged_event_grid` is called on `path_data`, so a perfectly dense slice is told its grid is ragged because of an asset in another slice. Measured with only `A11` (cohort `fast`) missing one period:

```
no price_data   dense 'slow' slice warning_codes=()
price_data      dense 'slow' slice warning_codes=('ragged_period_grid',)
```

### Suggested fix

Scope `price_data` to the slice's own assets before forwarding, e.g. filter it on `sub_df["asset_id"].unique()` inside the loop. Asset-level scoping is the right granularity: it keeps each asset's complete date range, so a path that crosses a slice boundary in time is still walked, and it needs no `by` column on `price_data`. Date-level scoping would undo the point of the PR — don't.

Measured with the price panel scoped that way, on the same panels:

| Slice | `baseline_bar_return` | `n_assets_in_baseline` | offset-6 excess |
|---|---|---|---|
| slow | 0.000500 | 6 | 0.0 |
| fast | 0.008000 | 6 | 2.2e-16 |

and the PR's actual goal is untouched — the tail offset is still restored either way, on the `offsets=[24]` panel from the new test:

```
first  unscoped         computed=6 censored=0 n_assets_in_baseline=12
first  scoped to slice  computed=6 censored=0 n_assets_in_baseline=6
```

If the panel-wide baseline is wanted deliberately, that is a defensible design, but it cannot flip silently on whether `price_data` was passed — it would need to be stated in the docstring and reachable without `price_data` too.

### Optional — the new test pins the leaking behaviour as its reference

`test_by_slice_routes_full_price_data_without_expanding_event_sample` asserts `restored_audit == direct_metric.metadata["per_offset"][24]`, where the direct call is given the same un-sliced `raw`. The two sides therefore agree *because* they share the defect, which is why the test passes. The consistency check is worth keeping — compare against `evaluate` with the slice-scoped price panel instead, and the assertion becomes the one that would have caught this.

Since #1052, `per_offset[k]` carries `benchmark` and `mean`, so that equality is now sensitive to the baseline and will need updating along with the fix.

### Not blocking

`by_slice` puts `price_data` after `by` / `factor_col` while `evaluate` puts it first among keyword arguments. Both are keyword-only, and `by` / `factor_col` are this function's identity, so the order reads fine as it is.

awwesomeman:
## Review findings fixed (`b1b78fde`)

Both required findings addressed at their shared cause, plus the optional one.

`price_data` is now restricted to each slice's own assets before it is forwarded. The restriction is **by asset, never by date**, so an asset keeps its whole price history in every slice it appears in and a path crossing a slice boundary in time is still walked. Coercion happens once for the whole panel; column and dtype validation stays with `evaluate`, which owns the price-panel contract and reports it against the slice it applies to.

**Baseline.** Re-measured on the same two-cohort panel, offsets `[6]`, events carrying no information:

| Call | Slice | `baseline_bar_return` | `n_assets_in_baseline` | offset-6 excess |
|---|---|---|---|---|
| no `price_data` | slow | 0.000500 | 6 | 0.0 |
| no `price_data` | fast | 0.008000 | 6 | 2.2e-16 |
| `price_data=raw` | slow | 0.000500 | 6 | 0.0 |
| `price_data=raw` | fast | 0.008000 | 6 | 2.2e-16 |

Passing prices no longer changes what a slice is measured against.

**Raggedness.** With only `A11` (cohort `fast`) missing one period, the dense `slow` slice reports `warning_codes=()` and `fast` reports `('ragged_period_grid',)`. The new guard asserts both halves, so it cannot pass by the warning disappearing everywhere.

**The feature itself is untouched.** The tail offset the parameter exists to restore still comes back: `computed=6`, `censored=0`, both before and after the scoping.

**The existing test's reference call is now scoped the same way.** It previously compared `by_slice` against `evaluate` holding the unsliced `raw`, so the two sides agreed because they shared the defect — that assertion now compares like with like and is the one that would have caught this.

### Against-unfixed evidence

Both guards were written first and run against `3687f6d5`:

```
FAILED tests/test_event_price_data.py::test_by_slice_baseline_is_formed_from_the_slice_not_the_price_panel
FAILED tests/test_event_price_data.py::test_by_slice_raggedness_is_measured_on_the_slice_not_the_price_panel
2 failed, 14 deselected
```

After the dispatcher change both pass, and `test_by_slice_routes_full_price_data_without_expanding_event_sample` failed on its stale unsliced reference (`benchmark` 0.10714 against 0.11511, `hit_rate` 0.5 against 0.0) until its reference call was scoped.

### Verification

Run on the branch with the repository interpreter:

- `ruff check` / `ruff format --check`: clean, 253 files
- `pytest`: **4079 passed, 44 skipped, 0 failed**
- `mkdocs build --strict`: built in 24.13s

Docstring, `docs/api/by-slice.md` and `factrix/llms-full.txt` all state the by-asset scoping and why the panel-level quantities have to stay inside the slice.

PR #1084, opened 2026-09-07
fix(event): withhold audits for withdrawn curves

### LINKED ISSUE #1077: fix(event): do not publish per_offset computed/censored counts on a withdrawn curve
## Problem

After PR #1064, the `event_around_return` `invalid_price_data` short circuit still publishes `per_offset[k] = {"computed": …, "censored": …}` counts for a curve whose output was withdrawn (`mean` is `None`). The counts describe work whose result was discarded, which is the "report what ran" exception CLAUDE.md asks to be documented explicitly rather than left implicit.

Found during review of #1064; not misleading on the value itself, so out of that PR's boundary.

## Expected

Either the short circuit publishes no per-offset counts, or the metadata states that the counts describe the discarded computation (documented exception, no sentinel reuse).

## Definition of Done

- [ ] Decide and document which of the two contracts applies.
- [ ] A test pins the short-circuit metadata shape and fails on the current state.
- [ ] `docs/api/metrics/event_horizon.md` states the exception.

Refs #1051 #1064

awwesomeman:
## Summary

- keep `per_offset` empty when an invalid baseline withdraws the entire event curve
- retain `n_events` / `n_obs` as the event count actually used by the raw-path computation
- document the baseline-withdrawal exception to the censor-audit contract
- add an against-unfixed regression for the exact short-circuit metadata shape

## Verification

- `uv run pre-commit run --all-files --hook-stage pre-commit`
- `uv run mypy factrix`
- `uv run pytest -p no:faulthandler tests/test_event_horizon.py tests/test_event_price_data.py tests/test_docs_stat_keys_by_metric.py tests/test_docs_examples.py -q` (156 passed, 44 skipped)

Closes #1077

### DIFF
--- a/docs/api/metrics/event_horizon.md
+++ b/docs/api/metrics/event_horizon.md
@@ -74,8 +74,13 @@ title: factrix.metrics.event_horizon
 !!! warning "Invalid prices withdraw the curve"
     `event_around_return` needs a finite unconditional bar-return baseline.
     Any observed non-finite or non-positive `price` invalidates that baseline,
-    so the metric returns `value=NaN`, audited `per_offset` entries, and
+    so the metric returns `value=NaN`, an empty `per_offset` mapping, and
     `WarningCode.METRIC_UNAVAILABLE` with `reason="invalid_price_data"`.
+    Raw paths may already have been formed before the baseline check, so
+    `n_events` / `n_obs` still report the distinct events that computation
+    actually used. Their per-offset computed/censored counts are deliberately
+    withheld because the curve they would describe was discarded. This is the
+    explicit short-circuit exception to the censor-audit contract below.
     Missing (`null`) observations remain allowed on ragged panels; fix invalid
     observed prices rather than interpreting a contaminated finite hit rate.
 
--- a/docs/reference/stat-keys-by-metric.md
+++ b/docs/reference/stat-keys-by-metric.md
@@ -510,9 +510,11 @@ Pre/post-event return profile; descriptive.
   negative offset cleared the 5-event floor — `0.0` there was the *best*
   possible score for a quantity never computed. An observed non-finite or
   non-positive price instead withdraws the entire curve with
-  `reason=invalid_price_data`, audited `per_offset` entries, and
+  `reason=invalid_price_data`, an empty `per_offset` mapping, and
   `WarningCode.METRIC_UNAVAILABLE`; `n_invalid_prices` reports the offending
-  row count. Null prices remain valid missing data on a ragged panel.
+  row count. `n_events` / `n_obs` still report the events used by the raw-path
+  computation, but its per-offset counts are withheld because that curve was
+  discarded. Null prices remain valid missing data on a ragged panel.
 
 - *benchmark horizon*: offset `k > 0` is a simple return from `t+1` to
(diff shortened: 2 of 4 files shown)


awwesomeman:
## Review — approve

No required findings. The change is correct, it matches an existing repo contract rather than inventing one, and it closes a code/docs divergence that predates the issue.

### The choice is the house rule, not a new exception

`docs/reference/stat-keys-by-metric.md` has said since #1040 (`d5b27a04`, "judge short-circuit metadata keys per key, not per branch") that a withheld result's keys are neutralised, and it names the neutralised form for each — `NaN` for `mfe_mae_ratio`, `None` for `baseline_bar_return`, **`empty` for `per_offset`**, `False` for `hac_applied`. It also names `n_events` in the opposite list, as a measurement taken before the bail that carries its real value.

So `per_offset={}` alongside a real `n_events` is exactly what that ledger already specified, and `main` was the side that disagreed: it published populated per-offset counts on a withdrawn curve while the contract page said the key should be empty. That divergence is older than #1077 and this PR fixes it too.

### The split against the other short circuit is principled

Both branches were checked on the branch:

```
no_price_data branch: reason=no_price_data      per_offset_keys=[6] n_obs=0
invalid_price branch: reason=invalid_price_data per_offset_keys=[]  n_obs=6 n_events=6
```

Keeping the audit in the first and withholding it in the second is not a per-branch judgement in disguise. In the `no_computed_event_offsets` / `no_price_data` branch every entry has `computed=0`, so the mapping asserts only "nothing was computed, here is why" — pure censoring information, and the only diagnostic that branch has. In the baseline branch `computed > 0`, so republishing those counts under the curve key would assert that six offsets' worth of curve survived when none did. The code comment says exactly that, which is the right place for it.

`n_events` is measured at line 293, before the baseline call at line 298, so "the raw paths ran" is a true statement about work that happened — verified: `n_obs=6`, `n_events=6` on the poisoned panel.

### Against-unfixed evidence

The renamed guard was run against `main`'s `event_horizon.py` with the rest of the branch in place:

```
FAILED tests/test_event_horizon.py::TestEventAroundReturn::test_zero_price_withholds_curve_and_its_per_offset_audit
E   AssertionError: assert {-6: {'eligib...{}, ...}, ...} == {}
E     'computed': 167,
E     'eligible': 169,
E     'mean': None, ...
1 failed, 29 deselected
```

It also strengthens the old assertion in the right direction: `assert result.metadata["per_offset"]` was a truthiness check that could not distinguish a populated audit from a correct one, and it is now a shape assertion paired with `n_obs > 0` and `n_events == n_obs`, so emptying the whole metadata would not make it pass.

### Verification reproduced

Run on the branch with the repository interpreter:

- `ruff check`: clean
- `pytest`: **4079 passed, 44 skipped, 0 failed**
- `mkdocs build --strict`: built in 24.82s

No consumer reads `per_offset` in a way an empty mapping breaks — the only subscripting sites are docs examples on valid panels.

### Non-blocking

The docstring now reads "Every **published** `per_offset[k]`", but the parallel sentence in `stat-keys-by-metric.md`'s *censor audit* bullet still reads "Every `per_offset` member reports …". It is vacuously true over an empty mapping, so nothing is wrong; the same qualifier would just make the two read alike if you touch that file again.

PR #1085, opened 2026-09-07
fix(inspect): count surviving common betas

### LINKED ISSUE #1074: fix(inspect): preflight marks common-scope beta metrics usable that evaluate refuses
## Problem

`inspect_data` can report `common_beta_profile`, `common_beta_r_squared` and
`common_beta_sign_consistency` as `usable` on a panel where `evaluate` refuses
them at run time with `n_assets=0`:

```
evaluate(): 3 metric(s) below their sample floor on this panel —
  common_beta_profile: insufficient_asset_beta_observations (n_assets=0, required=1);
  common_beta_r_squared: insufficient_asset_r_squared_observations (n_assets=0, required=1);
  common_beta_sign_consistency: insufficient_assets_for_sign_consistency (n_assets=0, required=2)
```

Reproduction (20 assets x 120 periods, factor broadcast per period, `forward_periods=5`):

```python
raw = fx.datasets.make_cs_panel(n_assets=20, n_dates=120)
data = compute_forward_return(
    raw.with_columns(pl.col('factor').mean().over('date').alias('macro')), forward_periods=5
)
info = fx.inspect_data(data, factor_cols=['macro'])
fx.evaluate(data, metrics=info.usable.to_metrics_dict(), factor_cols=['macro'], forward_periods=5)
```

**Corrected 2026-09-07 — measured, the trigger is not COMMON scope.** The
original wording above said this happens "on a broadcast COMMON-scope factor
column". Broadcasting is not what breaks it. `make_cs_panel` demeans its factor
cross-sectionally on every date except the last 6, and those 6 are exactly the
tail `compute_forward_return(forward_periods=5)` drops. So in the evaluation
panel every surviving date's broadcast mean is numerically zero: the `macro`
column's per-asset variance over time is **2.006e-33**, against `EPSILON =
1e-9`. `compute_common_betas` drops all 20 assets at its `factor_var > EPSILON`
gate, and `n_assets=0` at run time is **correct**.

Measured on the same panel before `compute_forward_return`, `macro` carries
real variation (per-asset variance 1.048e-03, max |macro| 0.2009) — it is the
tail-dropping that leaves only floating-point residue behind.

A genuinely time-varying broadcast factor is unaffected: it stays `usable` and
runs under `evaluate(strict=True)`. The defect is therefore entirely on the
preflight side, which compared the downstream asset floor against the raw
20-asset universe instead of the producer's surviving rows.

`spanning_alpha` is refused on the same call with `no_base_factors` — that one is a schema gate no preflight axis models, and is arguably out of scope for the same fix.

This is a preflight/run-time disagreement of the class `tests/test_inspect_evaluate_agreement.py` exists to catch, but that sweep builds only individual dense panels (`make_cs_panel` + `_panel()`), so the COMMON-scope cell is never swept. The verdict is the one users are told to trust: `inspect_data(...).usable.to_metrics_dict()` is the documented discovery bridge into `evaluate`.

Found while writing the multi-factor docs example in #1056 (the example uses `strict=False` to work around it); out of that PR's boundary.

## Expected

Either the preflight blocks these metrics on a common-scope panel with the reason the run-time floor states, or the run-time `n_assets` count is wrong and the metrics should run. Whichever it is, the two must agree.

## Definition of Done

- [x] Establish which side is wrong: settled by measurement above — `n_assets=0` is correct for this panel, and preflight is the side to fix.
- [ ] Fix that side so `inspect_data(...).usable.to_metrics_dict()` runs under `evaluate(strict=True)` on a COMMON x DENSE x PANEL panel.
- [ ] Extend the agreement sweep to the COMMON-scope cell so this class of disagreement cannot return silently.
- [ ] Decide separately whether `spanning_alpha`'s `no_base_factors` schema gate should be preflighted (it is currently exempted by the `no_*` out-of-scope rule).

awwesomeman:
## Summary

- make `inspect_data` apply common-beta asset floors to the per-asset regressions that actually survive `compute_common_betas`
- mirror the producer's finite-pair, minimum-history, and factor-variation filters in preflight
- use the surviving count for both hard blockers and `few_assets` warn tiers
- extend inspect/evaluate agreement coverage from INDIVIDUAL to COMMON dense panels and document the effective sample

## Root cause

The issue reproduction broadcasts each period's cross-sectional mean of an already demeaned synthetic factor. Its remaining time variation is only floating-point noise (about `1e-16`), so `compute_common_betas` correctly drops all assets at the `factor_var > EPSILON` gate. Inspect incorrectly compared downstream floors with the raw 20-asset universe instead of the producer's zero-row output.

A genuinely time-varying broadcast factor remains usable and runs under `evaluate(strict=True)`.

`spanning_alpha`'s `no_base_factors` result remains separate: it is a caller-supplied baseline/configuration schema gate, not a data-shape `SampleThreshold`, so this PR does not fold it into sample preflight.

## Verification

- regression test observed failing before the implementation
- `uv run pytest tests/test_inspect_evaluate_agreement.py tests/test_inspect_data.py tests/test_common_betas.py -q` — 275 passed
- three focused survivor-count cases — 3 passed
- `uv run ruff check .` — passed
- `uv run ruff format --check .` — 253 files already formatted
- `uv run mypy factrix` — no issues in 83 source files
- `git diff --check` — passed
- local strict docs build could not start without the docs extra; `uv run --extra docs` then stalled during dependency/inventory setup with no output and was stopped after 90 seconds, so CI should run the authoritative docs build

Closes #1074


awwesomeman:
## Review — approve

No required findings. I re-derived the root-cause claim rather than reading it, because the whole fix rests on which side of the disagreement is wrong, and it holds.

### The premise checks out, and it is sharper than the issue said

Measured on the issue's exact reproduction, `macro`'s per-asset variance over time is **2.006e-33** against `EPSILON = 1e-9`, so `compute_common_betas` drops all 20 assets at its `factor_var > EPSILON` gate and `n_assets=0` at run time is correct. Preflight was the wrong side.

What the issue did not say, and what I measured: **broadcasting is not the trigger.** `make_cs_panel` demeans its factor cross-sectionally on every date except the last 6 — and those 6 are exactly the tail `compute_forward_return(forward_periods=5)` drops:

```
dates with non-negligible cross-sectional mean: 6 of 120
  2020-04-25 .. 2020-04-30, |m| up to 0.2009
those dates survive compute_forward_return: 0
```

Before `compute_forward_return` the same column carries real variation (per-asset variance 1.048e-03, max |macro| 0.2009); after it, only floating-point residue survives. So the degeneracy is a property of this fixture's demeaning meeting the tail drop, not of COMMON scope. I have corrected the issue body to state that, since it was framing the trigger as "broadcast COMMON-scope factor column".

That distinction is what makes the fix safe rather than over-blocking, and this PR gets it right on both sides:

```
repro   usable common_beta_*: []                    evaluate(strict=True): OK
varying usable common_beta_*: [profile, r_squared,  evaluate(strict=True): OK
                               sign_consistency]
```

A genuinely time-varying broadcast factor stays usable and runs. On `main` the first line raises.

### The mirror is faithful

`_compute_common_beta_stage1_profile` reproduces the producer's three gates exactly — `_finite_expr(factor) & _finite_expr(return)`, `n_periods >= MIN_COMMON_BETA_PERIODS_HARD`, `var > EPSILON` — against `_common_betas.py:233-259`. The producer's final `beta.is_not_null() & is_finite()` filter needs no mirror: `beta` is null exactly when `_var_x <= EPSILON`, which the variance gate already covers, and finite elsewhere because the pair mask rules NaN out of the moments. Duplicating producer logic in preflight is a drift risk in principle, but it is the pattern already set by `_compute_ic_stage1_profile` and `_compute_fm_stage1_profile`, and the agreement sweep is what pins the two together.

### The warning-tier line change is correctly scoped

`cross_section_tier(properties.n_assets)` → `cross_section_tier(av.n)` reads wider than the PR describes, so I checked every override of `threshold_properties`: the IC and FM branches replace `n_periods` only (`_inspect.py:1181`, `:1213`), and the new common-beta branch is the sole `n_assets` override. So `av.n == properties.n_assets` for every other metric and the change is inert outside this cell — while inside it, the `few_assets` tier now describes the sample the metric will actually run on. That is the "report what ran" rule, not a drive-by.

### Against-unfixed evidence

Reproduced with `main`'s `_inspect.py` and the rest of the branch in place:

```
FAILED tests/test_inspect_evaluate_agreement.py::test_common_beta_preflight_uses_assets_surviving_the_producer
FAILED tests/test_inspect_evaluate_agreement.py::test_common_beta_preflight_warns_on_surviving_asset_count
2 failed, 152 passed
```

The sweep extension is the durable part: parametrising `_panel(common_factor=...)` over both scopes closes the cell the issue identified as never swept.

### Verification, including the gate you could not run

The docs extra resolves in this checkout, so I ran the build the PR body deferred to CI:

- `mkdocs build --strict`: **built in 25.54s**
- `ruff check` / `ruff format --check`: clean, 253 files
- `pytest`: **4142 passed, 44 skipped, 0 failed**

### On the deferred DoD item

Leaving `spanning_alpha`'s `no_base_factors` out is right and needs no follow-up issue: it is a caller-supplied configuration gate, not a data-shape `SampleThreshold`, so there is no preflight axis it could be modelled on. The PR body states that reasoning, which is what DoD item 4 asked for. It also does not surface on this reproduction — `spanning_alpha` is not in `usable` either way.

PR #1086, opened 2026-09-07
feat(multiple-testing): declare unsubmitted family members

### LINKED ISSUE #1072: feat(multiple-testing): accept a declared family size larger than the submitted results
## Problem

#1057 (PR #1071) made inactive-family accounting explicit: every submitted candidate counts toward `m` under `inactive_policy="count"`, and `family` metadata discloses `n_candidates_declared`, `n_tests_computed`, `n_inactive`, `n_tests_adjusted` and the policy. The third option recorded in #1057 — a declared research manifest whose family size exceeds the results actually submitted (candidates that never produced a `MetricResult` at all) — was deliberately left out because it does not fall out of that change.

## Expected

A user can state a plan-level family size (or manifest of candidate identities) so that `n_candidates_declared` reflects the pre-registered research plan even when some candidates never reached the screening verb.

## Open design points

- Input surface: an integer `family_size` versus an explicit manifest of identities that can be matched to submitted results.
- Interaction with `expand_over` buckets: whether the declared count is per bucket or total, and how it is split.
- Interaction with per-identity k-of-m denominators in `partial_conjunction` and the hierarchical verb.
- Reporting: the missing candidates must appear in `family` metadata as a separate count, not folded into `n_inactive`.

## Definition of Done

- [ ] One documented input surface, shared by all five screening verbs.
- [ ] `family` metadata separates declared-but-unsubmitted candidates from submitted-inactive ones.
- [ ] Tests cover declared > submitted, declared == submitted, declared < submitted (must raise), and the bucketed case.
- [ ] Docs extend the `m=1` vs `m=10` example in `docs/api/multi-factor.md`.

Refs #1057

awwesomeman:
## Summary

Adds a shared `family_size` declaration to all five high-level multiple-testing verbs so a pre-registered research family can remain larger than the results that reached screening.

- accepts a scalar per bucket / identity / group, or a strict mapping for heterogeneous sub-families
- rejects declared sizes below the submitted subset and mappings with missing or unexpected keys
- keeps declared-but-unsubmitted candidates as inert non-rejections under both inactive policies
- adds `FamilyAccounting.n_unsubmitted`, separate from submitted inactive candidates
- applies the declared denominator to flat/cross-metric BHY, k-of-m partial conjunction, and hierarchical inner BHY/Simes
- documents the `m=1` versus `m=10` effect and bucket semantics

## Statistical contract

`inactive_policy=exclude` only removes submitted candidates that are known to be inactive. It does not remove unsubmitted candidates, whose activity is unknown. For partial conjunction, unsubmitted conditions are explicit `p=1` non-rejections in the k-of-m denominator. For hierarchical screening, the declaration is per observed group and affects both its Simes representative and inner BHY denominator; an integer count cannot invent a wholly absent group identity.

## Verification

Against the unfixed implementation, the initial regression guard produced 5 failures: every call rejected the new kwarg. The final guard was expanded to six tests covering larger/equal/smaller declarations, bucket mappings, inactive-vs-unsubmitted accounting, all five verbs, and k-of-m numerics.

- `pytest` focused screening suite: 140 passed
- placeholder-policy + multiple-testing primitive suite: 121 passed before the warning-text expectation was updated; final placeholder/declared suite: 14 passed
- docs tests: 425 passed, 45 skipped
- module doctests: 97 passed, 1 skipped
- `ruff check .`: passed
- `ruff format --check .`: 254 files formatted
- `mypy factrix`: passed (83 source files)

Local Windows Polars printed its existing access-violation diagnostics during some successful pytest runs, but each reported command exited successfully. The base `.venv` lacks the Material docs theme; `uv run --extra docs mkdocs build --strict` then remained silent during environment setup/build and was stopped, so the strict site build is left to PR CI as agreed.

Closes #1072


awwesomeman:
## Review — approve

No required findings. The statistical contract is the hard part here and it holds under measurement, not just under reading.

### BY arithmetic checked against an independent implementation

I re-derived the step-up separately — pad the submitted p-values with `1.0` to the declared length, apply `min(1, m·c(m)·p₍ᵢ₎/i)` with `c(m) = Σ1/i`, then enforce monotonicity — and compared:

| declared `m` | factrix `adj_p_all` | independent reference | match |
|---|---|---|---|
| 2 | 0.030000 | 0.030000 | ✓ |
| 5 | 0.114167 | 0.114167 | ✓ |
| 10 | 0.292897 | 0.292897 | ✓ |

(two submitted p's, `0.01` and `0.02`). The `m=10` figure reproduces the 0.2929 the docs example publishes, from a different code path than the one that generated it.

**Monotonicity swept, not argued.** Every multiset of 1–3 p-values from a 5-point grid × declared `m` from `k` to `k+5` — **330 cells, 0 violations** of "raising the declared family size never lowers an adjusted p". That is the invariant a declared denominator has to satisfy to be conservative, and it is the one a formula argument would most easily get wrong at the step-up stage.

### The stated contract is the one implemented

The PR's key claim is that `inactive_policy="exclude"` removes *submitted* candidates known to be inactive but never *unsubmitted* ones, whose activity is unknown. Measured on one submitted result with `family_size=8`:

```
count   : family(declared=8, computed=1, inactive=0, unsubmitted=7, adjusted=8, policy='count')
exclude : family(declared=8, computed=1, inactive=0, unsubmitted=7, adjusted=8, policy='exclude')
```

`adjusted` stays 8 under both — the exclusion does not reach the unsubmitted, exactly as the Statistical contract section says. `n_unsubmitted` is separate from `n_inactive` rather than folded into it, which is what #1072's DoD asked for. Under-declaration is rejected: `family_size=1` against 3 submitted, and `0` / `-3`, all raise `UserInputError`.

### Verification you could not run locally

The PR body leaves the strict site build to CI and lists focused suites rather than a full run. This checkout resolves the docs extra, so I ran both:

- `mkdocs build --strict`: **built in 27.36s**
- `pytest` (full suite): **4149 passed, 44 skipped, 0 failed**

### Non-blocking

`docs/api/partial-conjunction-across-metrics.md`'s `to_frame()` comment still lists the pre-#1071 columns (`active`, `n_tests`) — stale before this PR, not caused by it, and already fixed in #1095. If #1095 merges first this will conflict trivially; whichever lands second just takes the other's version of that block.

PR #1103, opened 2026-09-07
fix(event): benchmark each asset on its own drift

### LINKED ISSUE #1088: fix(event): benchmark each asset on its own drift, not a pooled mean
## Problem

PR #1079 (#1052) made offset `k`'s benchmark `(1 + mean_mu)**k - 1`, where `mean_mu` is the equal-weighted mean of the per-asset single-period drifts (`_unconditional_bar_return` in `factrix/metrics/event_horizon.py`, around line 100; `_horizon_benchmark` around line 118). That is one pooled scalar, compounded once and subtracted from returns that are per-asset. On a panel whose assets drift at different rates the two do not cancel and a residue survives.

Measured on a deterministic drift-only panel: 8 assets compounding at 0.1% ... 1.5% per period (`np.linspace(0.001, 0.015, 8)`), 200 periods, events every 17 periods from period 40. No event carries information, so every excess return is zero by construction.

```
  k | published excess | benchmark applied |  n
 -3 |      +0.00016216 |       +0.00800000 | 74
 -1 |      +0.00016216 |       +0.00800000 | 74
  1 |      +0.00016216 |       +0.00800000 | 74
  6 |      +0.00032521 |       +0.04897030 | 72
 12 |      +0.00391940 |       +0.10033869 | 66
 24 |      +0.00560525 |       +0.21074524 | 63
```

Two distinct residues are in there:

- **A composition residue, flat across offsets.** Events are not spread equally over the assets (per-asset counts `[9, 9, 9, 9, 9, 9, 10, 10]`), so the event-weighted mean drift is `0.00816216` against the equal-weighted baseline of `0.00800000`. The gap `+0.00016216` reproduces the `k <= 1` rows exactly. Compounding is not involved here.
- **A Jensen gap that grows with `k`**, stacked on top: `+0.00032521` at 6, `+0.00391940` at 12, `+0.00560525` at 24.

This is second order relative to what #1079 fixed. The same panel with the pre-#1079 single-period benchmark (`_horizon_benchmark` returning `baseline` at every offset):

```
  k |  pre-#1079 |    merged | removed
  6 |  +0.041296 | +0.000325 |   99.2%
 12 |  +0.096258 | +0.003919 |   95.9%
 24 |  +0.208350 | +0.005605 |   97.3%
```

A per-asset benchmark prices to zero for any drift vector. Benchmarking each asset on its own compounded drift — measured by running each asset's sub-panel alone, so the baseline *is* that asset's own drift — gives a mean excess of at most `3e-17` at every offset, `k <= 1` included.

#1052's Definition of Done asked that "a deterministic drift-only panel produces approximately zero excess at all offsets". That now holds only for a *homogeneous* drift panel. The guard `_drift_only_panel` in `tests/test_event_benchmark_horizon.py` gives every asset the same `growth`, so it cannot see this: the homogeneous control at the same mean drift prices every offset at `<= 1e-16`.

## Expected

Compound each asset's own mean single-period drift and subtract it event-wise, so the benchmark matches both the horizon it spans and the asset the return came from. A drift-only panel then prices to approximately zero at every offset for any per-asset drift vector, not only a constant one.

## Definition of Done

- [ ] The benchmark is per-asset: each event's excess subtracts that asset's own drift compounded over the offset's own span in periods.
- [ ] `_drift_only_panel` gains a heterogeneous-drift variant; run it against the current state first and watch it fail (it reports `+0.0056` at `k=24` today).
- [ ] A case covers the composition residue at `k <= 1`, where compounding plays no part — a fix that only corrects the Jensen term leaves `+0.000162` there.
- [ ] `per_offset[k]["benchmark"]` keeps reporting the quantity actually subtracted; if that is no longer a single scalar, the key states what it publishes and the metadata does not contradict it.
- [ ] `docs/api/metrics/event_horizon.md` states the per-asset benchmark.

Refs #1052 #1079

awwesomeman:
## Summary

- estimate the unconditional single-period drift per asset and subtract the matching asset's horizon-compounded drift from every event
- report `per_offset[k][benchmark]` as the event-weighted mean signed benchmark actually subtracted, while retaining `baseline_bar_return` only as an equal-asset-weighted panel diagnostic
- publish `baseline_bar_return_by_asset` and a machine-readable `benchmark_weighting` contract
- censor event paths whose asset cannot form an adjacent-period baseline instead of borrowing another asset's drift
- update the event metric, slicing, stat-key, primitive, and LLM reference documentation

## Quantitative evidence

The two primary guards were written and run against the unfixed tree first:

```text
FAILED test_heterogeneous_drift_panel_has_no_excess_at_any_offset
  offset -3 keeps +0.000162162 of heterogeneous pure drift
FAILED test_single_bar_benchmark_follows_the_events_asset_mix
  reported benchmark 0.008000000; actual event-weighted benchmark 0.008162162
2 failed, 3 passed
```

After the fix, on 8 deterministic assets drifting from 0.1% through 1.5% per period, all event effects exactly zero:

| offset | mean excess | benchmark actually subtracted | n |
|---:|---:|---:|---:|
| -3 | -1.86e-17 | 0.0081621622 | 74 |
| -1 | +2.40e-18 | 0.0081621622 | 74 |
| 1 | -1.20e-17 | 0.0081621622 | 74 |
| 6 | -1.85e-17 | 0.0492955151 | 72 |
| 12 | -1.68e-17 | 0.1042580894 | 66 |
| 24 | +2.82e-17 | 0.2163504912 | 63 |

This removes both defects separately: the one-bar event-composition residue and the long-horizon Jensen/compounding residue.

## Verification

Executed locally:

- focused event and metadata-contract suite: **81 passed**
- docs contract suite: **291 passed**
- doctests: **97 passed, 1 skipped**
- Ruff check / format: clean, 254 files
- mypy: no issues in 83 source files

Not executed locally:

- complete pytest suite with the required development and documentation extras; the previously reported full-suite count came from an incomplete environment and has been removed
- `mkdocs build --strict` with the complete documentation environment; the local command could not start a complete docs build because the docs extra was absent

Closes #1088


awwesomeman:
## Review — approve

No required findings. The statistical claim is the whole PR and it reproduces from an independent recomputation, not just from the guard.

### The per-asset benchmark prices heterogeneous drift to zero

Rebuilt the panel from the issue (8 assets drifting 0.1% through 1.5% per period, events carrying no information, an uneven date grid):

| offset | mean excess | benchmark subtracted | n |
|---:|---:|---:|---:|
| -3 | -2.34e-20 | 0.0081621622 | 74 |
| -1 | -2.34e-20 | 0.0081621622 | 74 |
| 1 | +0.00e+00 | 0.0081621622 | 74 |
| 6 | +6.17e-18 | 0.0492955151 | 72 |
| 12 | -3.70e-17 | 0.1042580894 | 66 |
| 24 | -5.29e-17 | 0.2163504912 | 63 |

`baseline_bar_return_by_asset` recovers the declared drifts exactly (`A0`→0.001 … `A7`→0.015), and `baseline_bar_return` stays at 0.008, their equal-weighted mean, which is the diagnostic the PR says it now is. Both residues #1088 named are gone: the k≤1 composition effect (+0.000162) and the long-horizon compounding gap (+0.005605 at k=24).

### `benchmark` is a summary, and it says so

Each event is now measured against its own asset's compounded drift, so the per-offset `benchmark` is necessarily a summary rather than "the number subtracted". I recomputed it independently — the event-weighted mean of `(1 + g_asset)^k - 1` signed per event — and it matches the published value to the last digit at offset 6 (`0.04929551508553914` both ways). `benchmark_weighting = "event_weighted_mean_signed_per_asset"` states the contract machine-readably, which is what keeps this inside "report what ran" rather than beside it.

### No drift is borrowed

Probed an asset present only every third period alongside three dense names: it is censored (`missing_entry_price`), `n_assets_in_baseline` drops to 3, and `baseline_bar_return_by_asset` carries only the three that could form one. The new `missing_asset_baseline` / `no_event_asset_baselines` pair is registered in both `stat-keys-by-metric.md` and the metric page.

### Against-unfixed evidence — stronger than the body claims

The body reports "2 failed, 3 passed". Run against `origin/main`'s `event_horizon.py` + `_event_returns.py` with the tests kept, **4 fail**:

```
test_heterogeneous_drift_panel_has_no_excess_at_any_offset - offset -3 keeps 0.000162162
test_single_bar_benchmark_follows_the_events_asset_mix     - assert 0.008 == 0.008162162 ± 1e-12
test_per_offset_reports_the_benchmark_it_subtracted        - KeyError: 'baseline_bar_return_by_asset'
test_event_path_without_an_asset_baseline_is_censored      - assert 'no_pre_event…' == 'no_event_asset_ba…'
4 failed, 2 passed
```

So the censoring guard is regression proof too, not a pin.

### Two verification notes

**The full-suite figure in the body is wrong.** It reports `4093 passed, 51 skipped`. This branch already contains `7f77add0`, no test is deleted (three are added), and `main` is at 4153/44 — so the expected figure is 4156/44, which is exactly what I measure here. The 4093/51 run was against something else, most likely an environment missing the optional adapter deps (7 extra skips). Nothing is wrong with the code; the number just should not be read as this branch's.

**The docs build you could not start, run here:**

- `mkdocs build --strict`: **built in 26.85s**
- `pytest`: **4156 passed, 44 skipped, 0 failed**
- `ruff check`: clean

PR #1107, opened 2026-09-07
fix(metrics): reject unidentified pooled beta slopes (#1104)

### LINKED ISSUE #1104: fix(metrics): pooled_beta's singular-design branch discards a computable slope and fabricates p=1.0
## Problem

`factrix/metrics/fm_beta.py:560` — the `singular_pooled_design_matrix` short circuit in `_pooled_beta_driscoll_kraay` breaks the short-circuit contract on four axes at once. All four are the classes #835, #993 and #1013 were written to close.

Reproduction — one constant factor column, 40 dates x 5 assets, the same panel both times:

```python
panel = pl.DataFrame([
    {"date": d, "asset_id": f"A{a}", "factor": 1.0,
     "forward_return": float(rng.normal(0, 0.01))}
    for a in range(5) for d in dates          # 40 dates, rng = default_rng(7)
])

pooled_beta(driscoll_kraay=True)(panel, overlap_periods=1)
#   value=nan  stat=None  p_value=1.0  alternative='two-sided'
#   reason='singular_pooled_design_matrix'  warning_codes=('metric_unavailable',)

pooled_beta()(panel, overlap_periods=1)        # clustered, same data
#   value=-0.0006598097969293243  stat=None  p_value=None  alternative=None
#   warning_codes=('degenerate_variance',)
```

1. ~~**A computable point estimate is discarded.**~~ **Corrected — this claim was wrong, see the correction below.** The clustered branch does return a number off the same data, but that number is not an identified slope.
2. **`p_value=1.0` and `alternative='two-sided'` are fabricated** for a variance failure that is not a data shortage. `p_value=1.0` is the conservative default for shortages; here it states a test that never ran.
3. **The reason prefix is outside the vocabulary.** `_short_circuit_output` documents `insufficient_*` (data shortage) and `no_*` (missing input/config); `singular_pooled_design_matrix` is neither, so `_is_inactive_hypothesis` (`_multi_factor.py:1003`) does not recognise it. Under `inactive_policy="exclude"` this cell stays in the BHY family `m` as an active hypothesis sitting at p=1.
4. **The exception class is wrong.** `_is_sample_shortfall_reason` returns `False`, so `strict=True` raises `UserInputError` — "change the call" — while the clustered branch on the identical input raises nothing at all.

The twin at `fm_beta.py:881` (`np.linalg.lstsq` `LinAlgError` on a 2-column design) shares the reason string but is near-unreachable: the `_finite_expr` filter above it removes the NaN input that would provoke it.

## Correction (2026-09-07) — the reproduction is rank-deficient, so claim 1 is wrong

Measured while reviewing #1107. The reproduction pairs a constant factor with an
intercept, so the design has **rank 1 for 2 parameters** and no slope is
identified. The `-0.00066` the clustered branch returns is NumPy's minimum-norm
split of the fitted constant, not an estimate. Holding the returns fixed and
rescaling the constant factor from 1.0 to 2.0 moves it to `-0.00053` — a genuine
slope would halve to `-0.00033`, and this does neither, because the split
depends on the column norms rather than on anything in the sample:

```
factor=1.0: clustered value=-0.0006598097969293243
factor=2.0: clustered value=-0.0005278478375434597
rank of [intercept, constant factor] design: 1 of 2 params
```

So on **this** input the Driscoll-Kraay branch was right to withhold the value,
and the clustered branch was wrong to publish one. The estimate-preservation
rule still applies to the case this reproduction failed to construct: a
**full-rank** design whose covariance estimation alone fails.

Claims 2, 3 and 4 stand as written — a rank-deficient design is still not a
data shortage, so `p_value=1.0` with a `two-sided` label, a reason token outside
the vocabulary, and the asymmetric exception class remain defects independent of
whether the slope survives.

## Expected

A rank-deficient design withholds both the slope and the test. A covariance-estimator
failure on an identified design withholds the test and keeps the estimate. Both
travel under a reason token the family-accounting and exception-routing predicates
recognise.

## Definition of Done

- [ ] A rank-deficient design withholds the slope on **both** covariance paths, so neither exposes the minimum-norm coefficient.
- [ ] On a full-rank design whose covariance estimation fails, the slope survives and `stat` / `p_value` / `alternative` are withheld together.
- [ ] The reason token joins the documented vocabulary so `_is_inactive_hypothesis` and `_is_sample_shortfall_reason` route it deliberately rather than by default.
- [ ] `strict=True` behaviour is the same on both branches for the same input, or the difference is documented with its reason.
- [ ] A regression test builds the singular design and asserts the full result shape on both branches; run it against the unfixed state first.
- [ ] Check `fm_beta.py:881` under the same contract while the decision is fresh.

Found by a short-circuit sweep (79 sites enumerated) while auditing the closed issues of the #1062 series.

Refs #835 #993 #1013

awwesomeman:
## Summary

- reject rank-deficient pooled designs before either covariance path can expose NumPy's arbitrary minimum-norm coefficient
- preserve an identified pooled slope when only Driscoll-Kraay or clustered covariance inversion fails, while withholding `stat`, `p_value`, and `alternative`
- route all degeneracies through the existing `degenerate_variance` contract so BHY marks them inactive and strict mode remains consistent
- document the distinct unidentified-design and withheld-inference result shapes

## Quantitative correction to the issue premise

The issue reproduction uses a constant factor together with an intercept. Its design rank is 1 for 2 parameters, so the reported `-0.0006598097969293243` is not an identified slope; it is NumPy's minimum-norm split of the fitted constant. Rescaling the same constant factor from 1 to 2 changes that coefficient to `-0.0005278478375434594` without changing any information in the sample.

This PR therefore withholds the slope for the reproduced rank-deficient design. It still implements the issue's intended estimate-preservation rule for the valid case: when OLS is full rank and only covariance estimation fails, the identified slope survives and only inference is withheld.

## Verification

Executed locally:

- unfixed targeted regression: **3 failed as expected**
- fixed targeted regression: **3 passed**
- related tests: **149 passed**
- `ruff check .`: passed
- `ruff format --check factrix tests`: passed
- commit object inspection: no GPG signature and no `Signed-off-by` trailer

Not executed locally:

- complete pytest suite with the required development and documentation extras; the previously reported full-suite count came from an incomplete environment and has been removed
- `python -m mypy factrix`; the earlier statement that CI was authoritative was not a local result
- `mkdocs build --strict` with the complete documentation environment

Closes #1104

### DIFF
--- a/docs/api/metrics/fm_beta.md
+++ b/docs/api/metrics/fm_beta.md
@@ -69,6 +69,12 @@ title: factrix.metrics.fm_beta
     If a finite-sample two-way covariance is non-PSD, the OLS slope remains
     descriptive but its test is withheld rather than replaced by one-way
     inference.
+    A rank-deficient pooled design does not identify a slope: factrix returns
+    `value=NaN`, withholds `stat` / `p_value` / `alternative`, and records
+    `variance_status="singular_pooled_design_matrix"` plus
+    `degenerate_variance`. This differs from a covariance failure after a
+    full-rank fit, where the identified slope remains available and only the
+    test fields are withheld.
 
 -   __Cross-section-robust pooled inference__
 
--- a/docs/reference/stat-keys-by-metric.md
+++ b/docs/reference/stat-keys-by-metric.md
@@ -184,9 +184,10 @@ is emitted.
   the cluster count G < 3 the test is short-circuited with `stat =
   None`, `value = NaN`, and `p_value = 1.0`: with two clusters the sample
   cannot support a clustered estimate, so the algebraic slope is not
-  exposed as a usable pooled beta. A non-PSD two-way covariance is the
-  other withheld-inference regime and keeps `value` — see
-  `variance_status` below.
+  exposed as a usable pooled beta. Rank-deficient designs withhold the
+  unidentified slope and all test fields without fabricating a conservative
+  p-value. Covariance failures after a full-rank fit keep the identified
+  `value` but withhold the test — see `variance_status` below.
 - Sample size: `MetricResult.n_obs` (row count entering the test).
 - *descriptive*: `overlap_periods` (the injected evaluation-grid horizon) and
   `overlap_adjustment_applied`. The latter is `False` for clustered covariance,
@@ -202,6 +203,13 @@ is emitted.
   support a slope test. The pooled OLS slope remains available, but `stat` /
   `p_value` are withheld with `WarningCode.DEGENERATE_VARIANCE`; factrix does
   not substitute a one-way covariance.
+- *descriptive* (conditional): `variance_status` =
+  `"singular_pooled_design_matrix"` for a rank-deficient design or a covariance
+  inversion failure. A rank-deficient design also carries `design_rank` and
+  `n_parameters` and has no identified `value`; if OLS was full rank and only
+  covariance inversion failed, its identified slope remains available. Both
+  shapes withhold `stat` / `p_value` / `alternative` and carry
+  `WarningCode.DEGENERATE_VARIANCE`.
 - *descriptive* (Driscoll-Kraay path, `driscoll_kraay=True`):
   `se_method` (`"driscoll_kraay"`), `n_periods` (length of the
   cross-sectional score-sum series), and `driscoll_kraay_lags` (the
--- a/factrix/metrics/fm_beta.py
+++ b/factrix/metrics/fm_beta.py
@@ -62,6 +62,7 @@ from factrix.inference.series_mean import _persistent_array_beyond_horizon
 from factrix.metrics._base import MetricBase
 from factrix.metrics._decorators import metric
 from factrix.metrics._helpers import (
+    _degenerate_metric_output,
     _degenerate_test_fields,
     _enforce_min_floor,
     _finite_expr,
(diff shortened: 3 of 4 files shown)


awwesomeman:
## Review — one required fix, then approve

**The premise correction is right and I have applied it to #1104's body.** My issue's claim 1 was wrong, and this PR is correct to reject it.

I re-measured the refutation independently. Holding the returns fixed and rescaling the constant factor:

```
factor=1.0: clustered value=-0.0006598097969293243
factor=2.0: clustered value=-0.0005278478375434597
rank of [intercept, constant factor] design: 1 of 2 params
```

A genuine slope on a doubled regressor would halve to `-0.00033`. This does neither, because the minimum-norm split depends on the column norms rather than on anything in the sample. So on that input the Driscoll-Kraay branch was right to withhold and the clustered branch was wrong to publish — the opposite of what I filed. #1104's body now carries that correction, with claims 2–4 left standing since a rank-deficient design is still not a data shortage.

Keeping the estimate-preservation rule for the case my reproduction failed to construct is the right call, and both shapes come out as documented:

| input | value | p_value | alternative | variance_status | codes |
|---|---|---|---|---|---|
| rank-deficient, DK | `nan` | `None` | `None` | `singular_pooled_design_matrix` | `degenerate_variance` |
| rank-deficient, clustered | `nan` | `None` | `None` | `singular_pooled_design_matrix` | `degenerate_variance` |
| full-rank, 2 clusters | `nan` | `1.0` | `two-sided` | — | `metric_unavailable` |

Both covariance paths now agree on the rank-deficient design — the asymmetry #1104 opened on is gone — and routing through `degenerate_variance` puts it inside the contract BHY and strict mode already read, which is cleaner than minting a token for it.

## Required — the `lint` job is red

`mypy factrix` fails, and it is one line:

```
factrix\metrics\fm_beta.py:1065: error: Name "metadata" already defined on line 1025  [no-redef]
Found 1 error in 1 file (checked 83 source files)
```

`metadata: dict[str, object]` is annotated twice in the same function scope — once inside the `except np.linalg.LinAlgError` block at 1025, once at 1065 after it. The `except` path returns, so nothing is wrong at runtime; mypy sees both annotations in one scope. Dropping the `: dict[str, object]` from the second occurrence is enough — the first annotation already binds the name for the rest of the function.

The PR body notes mypy could not run in that virtualenv. This checkout has it, hence the exact error above. `ruff check` and `ruff format --check` are both clean (254 files).

## One verification note

The body reports `4,076 passed, 50 skipped`. This branch is on top of `ef6f064b` and adds tests, and `main` is at 4156/44 — measured here it is **4158 passed, 44 skipped, 0 failed**. The 4076/50 run was against a different environment (6 extra skips reads like the optional adapter deps missing). The code is fine; the number just should not be read as this branch's. Same pattern as #1103's body, so it may be worth checking what that other environment is before quoting its counts again.

## After the mypy fix

Nothing else blocks. Re-run `mypy factrix` and the `lint` job should go green; the rest of the matrix is already passing.

PR #1108, opened 2026-09-07
fix(inference): preserve p-value tail provenance

### LINKED ISSUE #1100: fix(monotonicity): signed_spearman_p_value is two-sided under a one-sided headline, unlabelled
## Problem

`factrix/metrics/monotonicity.py:477` publishes `signed_spearman_p_value` from `_p_value_from_t(t_signed, len(mono_arr))` — the two-sided default — while the result is built with `alternative="greater"` (line 493). No `*_alternative` key appears on the result, and `docs/reference/stat-keys-by-metric.md:563-566` lists the key under the descriptive Spearman shape without stating its tail.

Reproduction (seed 20260907, 120 dates x 40 assets, `forward_return = 0.05 * factor + 0.5 * noise`):

```
headline alternative = 'greater'   p_value = 0.561
metadata keys naming a tail: <none>
signed_spearman_t = 2.114138344338103   n_valid_periods = 24
signed_spearman_p_value = 0.04555535401500074
  two-sided recompute   = 0.04555535401500074   <- matches
  greater  recompute    = 0.02277767700750037   <- a factor of two apart
```

A reader who takes the headline's `alternative` as the study's tail reads this secondary p as one-sided and is wrong by a factor of two, in the conservative direction here but not by construction.

This is the pattern #991 fixed for `quantile_spread`'s `long_p_value` / `short_p_value`. It is weaker than that case — the signed Spearman is a different statistic rather than a decomposition of the headline — so the fix may well be the same one-key remedy rather than a change of tail.

## Expected

Every published p-value is labelled with the tail it was computed under, on the same result that publishes it.

## Definition of Done

- [ ] `signed_spearman_p_value` carries its tail in metadata, following #991's `leg_alternative` precedent, or is recomputed under the headline's alternative if that is the intended quantity — decide which and say why in the docstring.
- [ ] `docs/reference/stat-keys-by-metric.md` states the tail beside the key.
- [ ] A test pins the label against the arithmetic and fails if they diverge.
- [ ] Sweep the remaining secondary p-values for the same pattern while the decision is fresh.

Refs #991

### LINKED ISSUE #1102: test(inference): InferenceResult does not say which tail it ran
## Problem

#989 fixed the silent-typo half of its report: every `Inference.compute` member now rejects an unknown `alternative` with a `UserInputError`. Its second suggestion — 把 `alternative` 放進 `InferenceResult.metadata`，讓直接呼叫的結果自己說明跑了哪一尾 — was not implemented.

`InferenceResult` (`factrix/inference/_base.py:46`) carries `stat`, `p_value`, `metadata`, `warnings`, `estimate`, `n_obs`, and no tail field; no member writes one into metadata. Measured on a 200-period series through `NEWEY_WEST.compute`, the three tails give 0.026382 / 0.013191 / 0.986809 and the three results are otherwise indistinguishable.

`MetricResult` does enforce the pairing — `factrix/_results.py:75` refuses a `p_value` without its `alternative`, on the ground that a p-value without the alternative it was computed under is uninterpretable. The same reasoning applies to the direct-call research interface the `factrix.inference` module docstring advertises; it is only the metric layer that is currently protected.

## Expected

A p-value published by `Inference.compute` states the tail it was computed under, on the result that publishes it.

## Definition of Done

- [ ] `InferenceResult` carries the alternative that ran — a field, or a metadata key with one name shared by every member.
- [ ] All four members (`NON_OVERLAPPING`, `NEWEY_WEST`, `HANSEN_HODRICK`, `STATIONARY_BOOTSTRAP`) set it from the branch that ran, not from the argument as received.
- [ ] `docs/api/inference.md` states the key beside the `compute` signature.
- [ ] A test asserts the label matches the arithmetic for all three tails, and fails if a member hardcodes one.

Refs #989 #991

awwesomeman:
## Summary

- add `InferenceResult.alternative` and populate it from all four series-mean inference members
- verify every member's three tail labels against the p-value arithmetic
- add an executable direct-call inference API example and document the result contract
- label monotonicity's secondary signed-Spearman p-value as two-sided, distinct from the one-sided MR headline
- document the secondary tail and sweep the remaining secondary p-values for mismatched provenance

## Secondary p-value sweep

- quantile long/short leg tests already carry `leg_alternative=two-sided`
- Fama-MacBeth's uncorrected p-value is two-sided like its headline test
- common-quantile's Spearman p-value is two-sided like its headline Wald test
- monotonicity was the only secondary p-value whose tail differed from the headline without its own label

## Verification

Executed locally:

- #1102 regression against the unfixed tree: **12 failed as expected** (4 members x 3 tails)
- #1100 regression against the unfixed tree: **1 failed as expected**
- `ruff check .`: passed
- `ruff format --check factrix tests`: passed
- both issue commits inspected: no GPG signature and no `Signed-off-by` trailer

Not executed locally:

- complete pytest suite with the required development and documentation extras; the previously reported full-suite and affected-suite counts came from an incomplete environment and have been removed
- `python -m mypy factrix`; the earlier statement that CI was authoritative was not a local result
- `mkdocs build --strict` with the complete documentation environment

Closes #1100
Closes #1102

### DIFF
--- a/docs/api/inference.md
+++ b/docs/api/inference.md
@@ -1,0 +1,37 @@
+---
+title: factrix.inference
+---
+
+::: factrix.inference
+
+## Direct `compute` result
+
+Every series-mean member has the same direct-call shape:
+
+```python
+from datetime import date, timedelta
+
+import numpy as np
+import polars as pl
+from factrix.inference import NEWEY_WEST
+
+data = pl.DataFrame(
+    {
+        "date": [date(2024, 1, 1) + timedelta(days=i) for i in range(60)],
+        "value": np.linspace(-0.1, 0.2, 60),
+    }
+)
+result = NEWEY_WEST.compute(
+    data,
+    value_col="value",
+    overlap_periods=5,
+    alternative="greater",
+)
+assert result.alternative == "greater"
+```
+
+`InferenceResult.p_value` is always paired with
+`InferenceResult.alternative`, which is the validated `"two-sided"`,
+`"greater"`, or `"less"` tail passed to the statistical kernel. Read these
+two fields together; metadata contains method-specific diagnostics such as
+bandwidth, effective degrees of freedom, or bootstrap seed.
--- a/docs/api/metrics/monotonicity.md
+++ b/docs/api/metrics/monotonicity.md
@@ -36,6 +36,9 @@ title: factrix.metrics.monotonicity
     metadata, not the headline — $\mathbb{E}|\rho| > 0$ under $H_0$ by
     Jensen, so mean $|\rho|$ has an `n_groups`-dependent noise floor
     (0.67 / 0.42 / 0.27 at $K = 3 / 5 / 10$) that reads like evidence.
+    The accompanying `signed_spearman_p_value` is explicitly two-sided and
+    reports `signed_spearman_alternative="two-sided"`; it is separate from
+    the headline MR test's one-sided `alternative="greater"`.
 
 </div>
 
--- a/docs/reference/stat-keys-by-metric.md
+++ b/docs/reference/stat-keys-by-metric.md
@@ -573,8 +573,11 @@ frequency on the same panels is 5.0% / 5.0% / 4.0% at a nominal 5%.
   [Resampling knobs](statistical-methods.md#resampling-knobs)).
 - *descriptive Spearman shape*: `mean_abs_spearman` (magnitude, ≥ 0),
   `mean_signed` (direction consistency), `signed_spearman_t`,
-  `signed_spearman_p_value`. A high magnitude with a near-zero signed mean
-  still says the factor sorts returns but flips sign across dates.
+  `signed_spearman_p_value`, and `signed_spearman_alternative` (always
+  `"two-sided"`). This secondary test asks whether mean signed Spearman differs
+  from zero in either direction; it does not use the distinct headline MR
+  test's one-sided `alternative="greater"`. A high magnitude with a near-zero
+  signed mean still says the factor sorts returns but flips sign across dates.
 - *descriptive*: `n_valid_periods`, `n_groups`, `tie_ratio`, `tie_policy`.
 - `warning_codes` (conditional): `HIGH_TIE_RATIO` under
   `tie_policy="ordinal"`, `FEW_ASSETS` when the median per-period
--- a/factrix/inference/__init__.py
+++ b/factrix/inference/__init__.py
@@ -9,6 +9,10 @@ metric accept?" is answered per-metric (e.g. ``ic`` accepts
 
 The namespace is scoped to the **series-mean** family
 (``compute(data, *, value_col, overlap_periods, alternative="two-sided")``).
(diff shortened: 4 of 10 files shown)


awwesomeman:
## Review — approve

No required findings. Both halves do what their issues asked, and the tail labels hold against the arithmetic rather than against the argument.

### #1102 — every member's label matches the p it published

Ran all four members on one 200-period series with a positive mean, `overlap_periods=1`, seed 20260907:

| member | two-sided | greater | less | greater + less | 2 x min |
|---|---|---|---|---|---|
| `NON_OVERLAPPING` | 0.084063 | 0.042031 | 0.957969 | 1.000000 | 0.084063 |
| `NEWEY_WEST` | 0.153081 | 0.076540 | 0.923460 | 1.000000 | 0.153081 |
| `STATIONARY_BOOTSTRAP` | 0.124000 | 0.062000 | 0.963000 | 1.025000 | 0.124000 |
| `HANSEN_HODRICK` | 0.083294 | 0.041647 | 0.958353 | 1.000000 | 0.083294 |

**0 label mismatches** across all 12 cells. The two identities that would catch a hardcoded tail both hold: the three analytic members give `greater + less = 1.0` exactly, and every member gives `two-sided = 2 x min(one-sided)`. The bootstrap's 1.025 is the discrete resample distribution with the `+1` correction counted in both tails, not a defect — and it is the one member where a label could have been copied from the argument without the arithmetic following, so it is worth that the numbers move.

A mistyped `alternative` still raises `UserInputError`, so #989's guard is intact underneath this.

### #1100 — the secondary p is labelled, and it is genuinely the other tail

```
headline alternative = 'greater'          p_value = 0.101
signed_spearman_t = 10.304844091943075    n_valid_periods = 120
signed_spearman_p_value = 3.474935737514735e-18
signed_spearman_alternative = 'two-sided'
published matches the two-sided recompute: True
```

Labelling rather than recomputing is the right call here: the signed Spearman is a different statistic, not a decomposition of the MR headline, so forcing it onto the headline's tail would have changed what it measures. The sweep in the body checks out — quantile legs already carry `leg_alternative`, and the Fama-MacBeth and common-quantile secondaries are two-sided like their own headlines, so monotonicity really was the only one whose tail differed without a label.

### Verification you could not complete

The body reports the full suite at `4,092 passed, 50 skipped, 1 failed` from **before** the docs-example correction, and only subsets afterwards — so the final tree never had a green full run recorded. Closing that, plus the mypy the virtualenv could not provide:

- `mypy factrix`: **no issues in 83 source files** (clean, unlike #1107 on the same environment claim)
- `pytest` (final tree): **4175 passed, 44 skipped, 0 failed**
- `mkdocs build --strict`: built in 26.01s
- `ruff check` / `ruff format --check`: clean, 254 files

Note the count: 4175 against `main`'s 4156. The 4,092/50 figure in the body is from the same non-repository environment that produced #1103's 4093/51 and #1107's 4076/50 — three PRs in a row now. Whatever that venv is, it is missing the optional adapter dependencies (6 extra skips) and is a stale checkout besides; worth pinning down before quoting its counts again, since it is the one number a reviewer cannot reproduce.

PR #1109, opened 2026-09-07
fix(tradability): align cost and turnover contracts

### LINKED ISSUE #1089: fix(tradability): net_spread(estimated_cost_bps=None) bypasses validation through a reused sentinel
## Problem

In `factrix/metrics/tradability.py`, `_unpack_cost_inputs` guards the `estimated_cost_bps` domain check with a `None` test (around line 927):

```python
if estimated_cost_bps is not None:
    _validate_estimated_cost_bps(estimated_cost_bps, func_name=func_name)
```

`None` there means "the caller is `breakeven_cost`, which has no such argument" — the parameter defaults to `None` in the helper's signature for exactly that reason. But `net_spread` does have the argument, so a user's own `None` is swallowed by the same sentinel and reaches the arithmetic unvalidated:

```
>>> net_spread(gross_spread=0.001, turnover=0.5, estimated_cost_bps=None)
TypeError: unsupported operand type(s) for /: 'NoneType' and 'int'
```

raised from `factrix/metrics/tradability.py:1380`, `cost_drag = 4 * (estimated_cost_bps / 10000) * turnover / holding_periods`. #1054's Definition of Done asked for a structured `UserInputError` on an out-of-domain cost input; this path produces a bare `TypeError` from inside the kernel instead, naming neither the function, the field, nor the expected domain.

This is the sentinel-reuse case CLAUDE.md names: one value carrying both "not supplied by this caller" and "supplied as `None` by the user", so the branch that cannot satisfy the contract is silently the same branch as the one that need not.

## Expected

`net_spread(estimated_cost_bps=None)` raises `UserInputError` naming `estimated_cost_bps` and the expected domain, in the same shape as every other cost-input rejection. The "this caller has no such argument" case is distinguished from a user-supplied `None` by something other than a value the user can also pass.

## Definition of Done

- [ ] `_unpack_cost_inputs` no longer uses `None` to mean both things — e.g. a dedicated sentinel, or `net_spread` validating its own argument before the call.
- [ ] `net_spread(gross_spread=0.001, turnover=0.5, estimated_cost_bps=None)` raises `UserInputError`; a test pins it and fails on the current state with the `TypeError` above.
- [ ] `breakeven_cost` still takes no `estimated_cost_bps` and is unaffected — a case pins that too.
- [ ] `docs/api/metrics/tradability.md` states the `estimated_cost_bps` domain alongside the existing `turnover` / `gross_spread` ones.

Refs #1054

### LINKED ISSUE #1090: fix(tradability): cost algebra coerces bool and str scalars instead of rejecting them
## Problem

The cost-algebra validators in `factrix/metrics/tradability.py` coerce with `float(...)` and then test `math.isfinite(...)` / a range, with no type check. `_unpack_cost_inputs` (around line 912) does the coercion:

```python
turnover_value = float(
    turnover.value if isinstance(turnover, MetricResult) else turnover
)
```

so anything `float()` accepts passes the domain gate. Measured:

```
>>> breakeven_cost(gross_spread=0.001, turnover=True)
value = 12.5   metadata = {'gross_spread': 0.001, 'turnover': 1.0, 'holding_periods': 5}

>>> breakeven_cost(gross_spread=0.001, turnover="0.5")
value = 25.0   metadata turnover = 0.5

>>> breakeven_cost(gross_spread="0.001", turnover=0.5)
value = 25.0   metadata = {'gross_spread': 0.001, 'turnover': 0.5, 'holding_periods': 5}
```

`turnover=True` is read as a **full rotation** of both legs and priced as such — a plausible typo (`turnover=True` for a flag) that returns a finite, publishable number instead of an error. `net_spread(..., estimated_cost_bps=True)` goes further and reports the unconverted `True` under the `estimated_cost_bps` metadata key while pricing it as 1 bps.

The rest of factrix rejects `bool` explicitly. `_validate_overlap_periods` in `factrix/_data_input.py` documents the convention — "a positive `int`, `bool` rejected" — and implements it as `if not isinstance(declared, int) or isinstance(declared, bool)`; `_is_finite_number` in `factrix/metrics/_helpers.py` (line 1967) does the same, "A bool is an int to Python and a string is not a number at all". So this is a local break with the house convention rather than an open question.

## Expected

The cost-algebra scalars are type-checked before coercion, on the same terms as the rest of the library: a real number, `bool` rejected, `str` rejected. Each rejection raises `UserInputError` naming the field and the expected domain.

## Definition of Done

- [ ] `gross_spread`, `turnover` and `estimated_cost_bps` reject `bool` and non-numeric scalars with `UserInputError`, matching `_validate_overlap_periods` / `_is_finite_number`.
- [ ] Tests pin each rejection and fail on the current state, where they return `12.5` / `25.0` respectively.
- [ ] A `MetricResult` input whose `value` is a plain float keeps working unchanged — a case pins that the tightening does not reach the producer path.
- [ ] `docs/api/metrics/tradability.md` states the accepted scalar types.

Refs #1054

### LINKED ISSUE #1091: fix(tradability): notional_turnover's joint empty-leg filter contradicts the per-leg "on its own" reading
## Problem

`notional_turnover`'s empty-leg filter is joint across the two legs (`factrix/metrics/tradability.py`, around line 674):

```python
.filter(
    (pl.col("n_top") > 0)
    & (pl.col("n_bot") > 0)
    & (pl.col("n_top_prev") > 0)
    & (pl.col("n_bot_prev") > 0)
)
```

A rebalance is dropped when *either* leg is empty on *either* date, so a leg with a perfectly well-defined churn is discarded because the *other* leg had no book. The per-leg metadata is then computed over the surviving dates only, and understates the leg that did churn.

Measured. 10 assets, 21 periods, `n_groups=5`, `rebalance_lag=1`, a constant per-asset factor so membership never moves and every turnover is 0. One period mid-panel carries a single-asset cross-section (`_assign_quantile_groups` puts a lone name in group 0, so the top leg is empty there); the lone name is one that was not previously in the bottom leg.

```
thin=False: n_rebalances=20  mean_top=0.0  mean_bottom=0.0
thin=True : n_rebalances=18  mean_top=0.0  mean_bottom=0.0
```

Replicating the documented formula `1 - m / max(j, k)` per leg over the same 20 transitions shows what the two discarded rebalances carried:

```
transition  9->10: top_churn=None (undefined)  bot_churn=1.0 (well defined)
transition 10->11: top_churn=None (undefined)  bot_churn=1.0 (well defined)
```

So the bottom book actually rotated fully twice in 20 rebalances. Per-leg scoping gives `mean_bottom_turnover = 2.0 / 20 = 0.1`; the joint filter reports `0.0`.

The joint filter has a justification the PR never states: it is what keeps `value == (mean_top_turnover + mean_bottom_turnover) / 2` exact, because all three means are taken over one shared date set. Verified on `make_cs_panel(n_assets=80, n_dates=180)` at `rng` 0/1/2/7: the identity holds to `< 1e-12` in all 4 cells. Scoping the filter per leg would break it.

What is wrong is that the justification is unstated while the docstring makes the opposite claim. The Returns section says `mean_top_turnover` is "the membership churn of the equal-weight top-quantile book **on its own** — the matched turnover proxy for a long-only allocation that holds the top bucket against a benchmark, which pays nothing for bottom-leg churn". A reader taking the long-only reading gets a number conditioned on the other leg existing. The Notes section does state the filter ("A rebalance is skipped when *either* date leaves *either* leg empty"), but does not connect it to the per-leg keys or to the identity it protects.

## Expected

One of two, chosen deliberately and documented:

1. **Scope the filter per leg.** `mean_top_turnover` / `mean_bottom_turnover` are each averaged over the rebalances where *that* leg was defined, making the "on its own" reading true. `value` then keeps the joint filter and stops being the mean of the two published per-leg keys, so the Returns section must say so — the identity is currently implied by "`value` is their mean".
2. **Keep the joint filter and drop the "on its own" phrasing.** State that the per-leg keys are conditioned on both legs existing on both dates, and state the identity that condition buys.

## Definition of Done

- [ ] Decide between the two and record the reason in the docstring, not only in a commit message.
- [ ] `mean_top_turnover` / `mean_bottom_turnover` and `value` each report the sample actually averaged over; if the two differ, the metadata says which rebalance count belongs to which key rather than publishing one `n_rebalances` for both.
- [ ] A test pins the thin-cross-section case above and fails on the current state (`mean_bottom_turnover == 0.0` where the bottom book rotated twice).
- [ ] `docs/api/metrics/tradability.md` matches whichever contract is chosen.

Refs #1053

### LINKED ISSUE #1099: fix(metrics): _lag_within_asset lags asset rows while its callers document periods
## Problem

`factrix/metrics/_helpers.py::_lag_within_asset` shifts each asset's own surviving rows. On a ragged panel that is not the quantity its callers document.

`factrix/metrics/quantile.py:668` states the contract as shifting `weight_col` by **1 period** per asset, and `_lag_within_asset`'s own docstring says "the value observed one sampled **period** earlier". Both name a period count the code does not apply when an asset is missing a period.

Reproduction (8-period grid, date strides 1 and 17, assets A and B, B missing grid period 3, `market_cap = 1000 * (i + 1)`): B's lagged weight at grid period 4 is taken from grid period **2** — a gap of 2 — while every other row's gap is 1.

Unlike the sibling defect in `rank_turnover` (fixed in #1098), the economics here are arguably fine: a stale weight is still look-ahead-free and is what a live book would be holding through a missing print. The defect is that the code and the sentence describing it name different quantities, so a reader cannot tell which one a ragged panel got.

## Expected

The lag is a count of periods on the panel's distinct-date grid, or the docstrings say rows and state the ragged-panel exception explicitly.

## Definition of Done

- [ ] Decide between densifying before the shift and documenting the row semantics; record why in the docstring.
- [ ] `_lag_within_asset` and the `quantile.py` contract sentence agree with the implementation and with each other.
- [ ] A ragged-panel test pins the chosen semantics, and fails on the other one.
- [ ] If the row semantics are kept, the exception is documented where a caller reads it, not only in the private helper.

Refs #1 #872 #1098

awwesomeman:
## Summary

- distinguish an invalid `estimated_cost_bps=None` from the sibling helper's absent argument
- reject boolean and string cost-algebra scalars before numeric coercion
- compute each turnover leg on its own valid rebalance sample while retaining a joint sample for the long-short headline
- lag value weights on the sampled panel's distinct-period grid so ragged assets cannot borrow stale weights across holes
- document the scalar domains, per-leg sample counts, and period-grid lag contract

## Quantitative contract

`notional_turnover` now reports three explicit sample sizes:

- `n_rebalances`: transitions where both legs exist, used by the long-short headline
- `n_top_rebalances`: transitions used by `mean_top_turnover`
- `n_bottom_rebalances`: transitions used by `mean_bottom_turnover`

The per-leg means are no longer forced onto the joint sample. Consequently, the headline equals the arithmetic mean of the two published leg means only when their samples coincide.

## Period-grid lag: what it costs

`_lag_within_asset` now shifts on the sampled panel's distinct-date grid, so
an asset absent at period *t* also loses *t+1* — its exact prior-grid weight
does not exist and no older observation is substituted. A stale carried weight
would be look-ahead-free as well, but it is a longer, asset-dependent lag; the
grid rule keeps every asset on the rebalance stride and matches the choice
`rank_turnover` made in #1098. The trade-off and the row loss are now recorded
where a caller reads them (`quantile_spread_vw` docstring, `docs/api/metrics/quantile.md`).

Measured on the deliberately ragged `_spread_panel` fixture in
`tests/test_drop_rate_warning_phase2.py` (3 assets on even periods, 50 on odd):

| | main | this branch |
|---|---|---|
| rows surviving the lag | 5300 → 5250 | 5300 → 597 |
| assets surviving the lag | 50 | 3 |
| result | 99 periods, `drop_rate` 0.50 | `metric_unavailable`, `insufficient_assets_for_quantile_groups` |

That fixture exists to pin the drop-stat schema on empty quantile legs, not to
exercise weight lagging, so it now passes `lag_weights=False` and keeps testing
what it was written for. Callers on ragged panels should read `median_cross_section` on successful
results or `max_assets_per_date` on an insufficient-assets short-circuit.
`drop_rate` separately counts period-level spreads lost after bucketing; it
does not count asset rows removed by the lag. Callers can also pre-lag their
weights and pass `lag_weights=False`.

## Verification

Executed locally (detached worktree on this branch, project venv):

- `ruff check .` and `ruff format --check .`: clean, 254 files
- `python -m pytest -q`: **4192 passed, 44 skipped, 0 failed**
- `python -m pytest --doctest-modules factrix/metrics/quantile.py`: 4 passed

Closes #1089
Closes #1090
Closes #1091
Closes #1099

### DIFF
--- a/docs/api/metrics/quantile.md
+++ b/docs/api/metrics/quantile.md
@@ -51,6 +51,10 @@ title: factrix.metrics.quantile
     diagnostics as its equal-weighted sibling, so the comparison is
     like-for-like rather than one leg being quietly less guarded.
 
+    The lag is one period on the sampled panel's distinct-date grid, within
+    each asset. If an asset is absent at the exact preceding grid period, its
+    current row has no lag and drops; an older stale weight is not substituted.
+
 -   __Per-bucket mean returns for monotonicity charts__
 
     ---
--- a/docs/api/metrics/quantile.md
+++ b/docs/api/metrics/quantile.md
@@ -55,6 +55,15 @@ title: factrix.metrics.quantile
     each asset. If an asset is absent at the exact preceding grid period, its
     current row has no lag and drops; an older stale weight is not substituted.
 
+    A stale weight carried across the hole would be look-ahead-free too, but
+    it is a longer, asset-dependent lag; one grid period keeps the weight lag
+    equal to the rebalance stride and matches `rank_turnover`. The cost is
+    rows: an asset absent at period *t* also loses *t+1*, so a ragged panel
+    whose thin periods alternate with full ones can drop most of its sample
+    and short-circuit as `metric_unavailable`. Read `drop_rate` and
+    `max_assets_per_date` to see what survived, or pass `lag_weights=False`
+    when the weights are already lagged.
+
 -   __Per-bucket mean returns for monotonicity charts__
 
     ---
--- a/docs/api/metrics/quantile.md
+++ b/docs/api/metrics/quantile.md
@@ -60,9 +60,12 @@ title: factrix.metrics.quantile
     equal to the rebalance stride and matches `rank_turnover`. The cost is
     rows: an asset absent at period *t* also loses *t+1*, so a ragged panel
     whose thin periods alternate with full ones can drop most of its sample
-    and short-circuit as `metric_unavailable`. Read `drop_rate` and
-    `max_assets_per_date` to see what survived, or pass `lag_weights=False`
-    when the weights are already lagged.
+    and short-circuit as `metric_unavailable`. A successful result reports the
+    post-lag breadth as `median_cross_section`; an insufficient-assets
+    short-circuit reports `max_assets_per_date`. `drop_rate` is narrower: it
+    counts period-level spread values lost after bucketing, not asset rows
+    removed by the lag. Pass `lag_weights=False` when the weights are already
+    lagged.
 
 -   __Per-bucket mean returns for monotonicity charts__
 
--- a/docs/api/metrics/tradability.md
+++ b/docs/api/metrics/tradability.md
@@ -129,9 +129,9 @@ float through the formula.
 
 | Input | Domain | Why |
 |---|---|---|
-| `gross_spread` | finite | A non-finite spread has no reading as a per-period return. |
-| `turnover` | finite, $0 \le \tau \le 1$ | The one-way per-leg replaced fraction above. `rank_turnover` lives in $[0, 2]$ and does not belong here. |
-| `estimated_cost_bps` | finite, $\ge 0$; `None` is invalid | A one-way per-trade cost. Omit the argument to use `net_spread`'s 30 bps default; `None` is not a default sentinel. A negative cost would make trading a source of return. |
+| `gross_spread` | finite real numeric scalar; `bool` / strings rejected | A non-finite spread has no reading as a per-period return. |
+| `turnover` | finite real numeric scalar, $0 \le \tau \le 1$; `bool` / strings rejected | The one-way per-leg replaced fraction above. `rank_turnover` lives in $[0, 2]$ and does not belong here. |
+| `estimated_cost_bps` | finite real numeric scalar, $\ge 0$; `None`, `bool` and strings rejected | A one-way per-trade cost. Omit the argument to use `net_spread`'s 30 bps default; `None` is not a default sentinel. A negative cost would make trading a source of return. |
 | `holding_periods` | integer $\ge 1$ | A rebalance interval in underlying return periods. |
 
 A violation raises `UserInputError`, with the same bounds applied to a bare
--- a/docs/api/metrics/tradability.md
+++ b/docs/api/metrics/tradability.md
@@ -114,9 +114,13 @@ not two approximations of each other.
     departed name and double the survivor: $\tau = 0.5$, where the $1 - m/k$
     reading was $0$.
 
-    A rebalance is skipped when *either* date leaves *either* leg empty: a
-    weight change needs both portfolios to exist. `metadata["n_rebalances"]`
-    counts the rebalances actually priced, and `mean_tail_size` /
+    A leg's churn is skipped only when that leg is empty on either date. Its
+    valid sample is reported as `metadata["n_top_rebalances"]` or
+    `metadata["n_bottom_rebalances"]`; an empty opposite leg does not erase a
+    well-defined long-only diagnostic. The headline long-short `value` still
+    needs both legs and `metadata["n_rebalances"]` counts that joint sample.
+    Consequently, `value` equals the arithmetic mean of the two published
+    per-leg means only when their samples coincide. `mean_tail_size` /
     `mean_top_tail_size` / `mean_bottom_tail_size` report the **current** leg
     sizes at $t$ — they equal the turnover denominator only while the legs do
     not shrink.
--- a/docs/api/metrics/tradability.md
+++ b/docs/api/metrics/tradability.md
@@ -131,7 +131,7 @@ float through the formula.
 |---|---|---|
 | `gross_spread` | finite | A non-finite spread has no reading as a per-period return. |
 | `turnover` | finite, $0 \le \tau \le 1$ | The one-way per-leg replaced fraction above. `rank_turnover` lives in $[0, 2]$ and does not belong here. |
-| `estimated_cost_bps` | finite, $\ge 0$ | A one-way per-trade cost. A negative cost would make trading a source of return. |
+| `estimated_cost_bps` | finite, $\ge 0$; `None` is invalid | A one-way per-trade cost. Omit the argument to use `net_spread`'s 30 bps default; `None` is not a default sentinel. A negative cost would make trading a source of return. |
 | `holding_periods` | integer $\ge 1$ | A rebalance interval in underlying return periods. |
 
 A violation raises `UserInputError`, with the same bounds applied to a bare
--- a/docs/reference/stat-keys-by-metric.md
+++ b/docs/reference/stat-keys-by-metric.md
@@ -1195,10 +1195,14 @@ inference.
 
 #### `notional_turnover`
 
-- *descriptive*: `n_rebalances`, `n_groups`, `overlap_periods`,
-  `rebalance_lag`, `mean_top_turnover`, `mean_bottom_turnover`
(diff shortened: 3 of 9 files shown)


awwesomeman (commented):
Verified locally in a detached worktree at 25875249 with `uv sync --all-extras`: `ruff check` clean, `mypy factrix` clean (83 files), `pytest tests/test_tradability.py tests/test_quantile.py` → 170 passed. The #1089 / #1090 / #1091 logic reads correctly: `None` is rejected at the public boundary, `bool` / `str` never reach `float()`, per-leg means run on their own samples while the headline keeps the joint one, and the three sample counts are the ones the code used.

**Required**

1. `ruff format` fails on three files (this is what turns the CI `lint` job red): `factrix/metrics/_helpers.py:1929`, `factrix/metrics/tradability.py:706`, `tests/test_quantile.py:286`.

2. CI pytest fails on every cell for a test outside this PR's files: `tests/test_drop_rate_warning_phase2.py::test_quantile_spread_vw_records_schema`. The grid-densified `_lag_within_asset` changes the outcome of that fixture (`_spread_panel`: even periods carry 3 assets, odd periods 50). Measured on the fixture:

   | | main | this branch |
   |---|---|---|
   | rows after lag | 5300 → 5250 | 5300 → 597 |
   | assets surviving the lag | 50 | 3 |
   | result | 99 periods, `drop_rate` 0.50 | `metric_unavailable`, `insufficient_assets_for_quantile_groups`, `max_assets_per_date=3` |

   Under grid semantics an asset absent at period *t* also loses period *t+1*, because its exact prior-grid weight does not exist. #1099's Definition of Done asked for a decision between densifying and documenting row semantics *with the reason recorded*; the issue itself notes a stale weight is look-ahead-free and economically defensible. The docstrings now state the grid rule but nowhere records why it was preferred over the stale-weight reading, and the PR body does not mention that ragged panels lose rows. Either choice can be right — #1098 chose the grid for `rank_turnover`, so consistency favours this one — but the trade-off has to be written down where a caller reads it, the data-loss consequence stated, and the broken test brought in line with the chosen semantics (the fixture as written is a pathological ragged panel; if grid semantics stay, the test needs a panel whose thin periods do not evict every full-period asset).

**Optional**

- `_validate_turnover` casts `value` twice (`float(cast(...))` in the guard and again in the return). Resolving once into a local reads cleaner; not blocking.

PR #1110, opened 2026-09-07
fix(compare): preserve data and deterministic ranks

### LINKED ISSUE #1092: fix(compare): a params key named __factrix_sort_key is silently overwritten and dropped
## Problem

`_rank` in `factrix/_compare.py` attaches its sort key under a fixed column name and drops it afterwards:

```python
_SORT_KEY = "__factrix_sort_key"
...
data = data.with_columns(key.alias(_SORT_KEY))   # line 274
...
return data.drop(_SORT_KEY)                      # line 292
```

`params` keys become output columns, so a caller whose `params` carries that name has their column overwritten by the internal key and then dropped with it. No error, no warning.

Measured, two results with `params={"__factrix_sort_key": 1}` / `{"__factrix_sort_key": 2}`:

```
>>> compare(results, metrics=["ic"], sort_by="ic_p_value").columns
['factor', 'forward_periods', 'ic', 'ic_p_value', 'rank']
```

The params column is gone. The control with the same params under a benign name keeps it:

```
>>> compare(results_ok, metrics=["ic"], sort_by="ic_p_value").columns
['factor', 'forward_periods', 'seed', 'ic', 'ic_p_value', 'rank']
```

Low severity — the name is obscure enough that nobody hits it by accident — but it is silent data loss in a documented output contract: the Returns section promises "params keys (union across results, first-seen order)". It is also reachable in the other direction: `sort_by="__factrix_sort_key"` names a column the caller can see in the input but not in the output.

## Expected

A `params` key colliding with the internal sort key is either rejected with `UserInputError` naming the reserved name, or the internal key is made non-collidable (a name derived so it cannot exist in the frame). Either way no user column is dropped without being told.

## Definition of Done

- [ ] `compare` no longer silently drops a `params` column named `__factrix_sort_key`.
- [ ] A test pins the chosen behaviour and fails on the current state, where the column is absent from `df.columns`.
- [ ] The `sort_by="__factrix_sort_key"` path behaves consistently with the choice.
- [ ] If the name is reserved, `docs/api/compare.md` says so where the params-column contract is stated.

Refs #1060

### LINKED ISSUE #1093: design(compare): order-invariance is not total for non-sortable params dtypes
## Problem

`compare`'s deterministic tiebreak covers only columns of a sortable dtype. `_sortable_tiebreaks` (`factrix/_compare.py`, line 251) keeps a column only when

```python
dtype.is_numeric() or dtype.is_temporal() or dtype in (pl.Boolean, pl.String)
```

so rows distinguishable *only* by a `params` column of any other dtype are equal on every tiebreak key and keep input order. Measured with 4 results identical except for a list-valued param (`{"window": [1, 2]}` ... `{"window": [7, 8]}`, dtype `List(Int64)`), swept over all input permutations:

```
input permutations:                    24
distinct output orderings:             24
rows that can receive ordinal rank 1:   4  ('[1, 2]', '[3, 4]', '[5, 6]', '[7, 8]')
```

Every permutation gives a different answer, and which row is ranked 1 follows the caller's list order. The control with the same shape under a sortable dtype (`window` as an `int`) gives 1 distinct ordering over the same 24 permutations.

The behaviour itself is documented accurately in the docstring's "Row order and ties" paragraph — "any `params` column of a sortable dtype", then "Rows equal on every one of those columns are indistinguishable and keep input order among themselves". Two things sit against it:

- The sentence between those two overstates. Docstring: "The output therefore does not depend on the order of `results`." `docs/api/compare.md` line 83 is stronger still, making it a consequence: "any `params` column of a sortable dtype, **so** the output does not depend on the order of the input `results`". Both are unconditional; the qualifier is only in the sentence after.
- The parameter table where a caller actually looks says nothing about it. `docs/api/compare.md` line 60 describes `rank_method="ordinal"` as "every row numbered `1..N`" with no pointer to the tie rules at all.

## Expected

A decision between two resolutions, both legitimate:

1. **Extend the tiebreak** to a stable representation of non-sortable params (a canonical string form, say), making order-invariance total and the unqualified sentence true.
2. **State the limit.** Keep the tiebreak as it is, qualify the "does not depend on the order of `results`" sentence in both the docstring and `docs/api/compare.md`, and say it at `rank_method` where a caller reading only the parameter table will see it.

Under (2), consider whether a `params` column that cannot participate in the tiebreak deserves a warning — a caller who passes list-valued params gets a silently order-dependent leaderboard either way.

## Definition of Done

- [ ] Choose between extending the tiebreak and documenting the limit, and record the reason.
- [ ] A test sweeps every permutation of a fixture whose rows differ only in a non-sortable `params` column and asserts the chosen contract; under (1) it must fail on the current state, which yields 24 orderings over 24 permutations.
- [ ] The docstring's "does not depend on the order of `results`" sentence and `docs/api/compare.md` line 83 agree with the choice.
- [ ] `rank_method`'s row in the `docs/api/compare.md` parameter table points at the tie rules.

Refs #1060

awwesomeman:
## Summary

- derive a non-colliding internal sort-key name so a params column named `__factrix_sort_key` remains visible and sortable
- canonicalize list, struct, and other non-native sortable values into internal type-tagged ordering keys
- keep caller-visible params values unchanged while making ordinal ranks invariant to input permutations
- link the `rank_method=ordinal` parameter contract to the deterministic tie rules

## Determinism contract

Native numeric, temporal, boolean, and string columns keep their Polars ordering. Other dtypes use a canonical type-tagged representation only for sorting; the original output column is preserved. Rows equal on every visible output column remain indistinguishable.

## Ordering caveat

The canonical key orders lexicographically on the tagged string, not on the
value's own semantics: measured, `_canonical_sort_value([10, 2])` sorts before
`_canonical_sort_value([2, 1])`. The docstring now says so, so callers read
such an order as stable rather than meaningful.

## Verification

Executed locally (detached worktree on this branch, project venv):

- `ruff check .` and `ruff format --check .`: clean, 254 files
- `python -m pytest -q`: **4179 passed, 44 skipped, 0 failed**
- `python -m pytest -k compare`: 54 passed, 1 skipped

Closes #1092
Closes #1093

### DIFF
--- a/docs/api/compare.md
+++ b/docs/api/compare.md
@@ -57,7 +57,7 @@ The returned `pl.DataFrame` contains the following columns:
 | `metrics` | (required) | `list[str]` of metric labels to include in the leaderboard. |
 | `sort_by` | `None` | Any output column produced before ranking: `factor`, `forward_periods`, `params` keys, metric value columns such as `ic`, or p-value columns such as `ic_p_value` / `<metric_label>_p_value`. `None` keeps the original list order. |
 | `descending` | `None` | Sort direction for `sort_by`. `None` resolves it from the column under the [direction rule](#sort-direction); `True` / `False` states it outright and always wins. There is no global direction default. |
-| `rank_method` | `"min"` | How equal `sort_by` values are numbered: `"min"` (tied rows share the best rank, next rank skips: `1, 1, 3`), `"dense"` (share the rank, no gap: `1, 1, 2`), `"ordinal"` (every row numbered `1..N`). Polars' `Expr.rank` methods. |
+| `rank_method` | `"min"` | How equal `sort_by` values are numbered: `"min"` (tied rows share the best rank, next rank skips: `1, 1, 3`), `"dense"` (share the rank, no gap: `1, 1, 2`), `"ordinal"` (every row numbered `1..N`, using the deterministic [tie order](#ties-order-and-missing-values)). Polars' `Expr.rank` methods. |
 
 ### Sort direction
 
@@ -80,10 +80,12 @@ signed metrics such as `predictive_beta`, `fm_beta` and `spanning_alpha`, where
 ### Ties, order, and missing values
 
 Rows sort on `sort_by`, then on `factor` and `forward_periods` ascending, then on
-any `params` column of a sortable dtype, so the output does not depend on the
-order of the input `results`. Rows equal on every one of those columns are
-indistinguishable and keep input order among themselves; under `"min"` and
-`"dense"` they carry the same rank anyway.
+the remaining params and output columns, so the output does not depend on the
+order of the input `results`. List, struct and other columns Polars cannot sort
+natively use a canonical, type-tagged string as an internal ordering key; the
+returned column and its values stay unchanged. Rows equal on every one of those
+columns are indistinguishable and keep input order among themselves; under
+`"min"` and `"dense"` they carry the same rank anyway.
 
 A `null` or `NaN` `sort_by` value sorts **last** in both directions and carries a
 `null` rank under every `rank_method` — a row with no value has no place in the
--- a/factrix/_compare.py
+++ b/factrix/_compare.py
@@ -267,29 +267,39 @@ def _rank(
     data: pl.DataFrame, *, sort_by: str, descending: bool, rank_method: RankMethod
 ) -> pl.DataFrame:
     """Sort ``data`` on ``sort_by`` and attach the ``rank`` column."""
+    sort_key_name = _unused_column_name(data.columns, _SORT_KEY)
     key = pl.col(sort_by)
     if data.schema[sort_by] in (pl.Float32, pl.Float64):
         # Fold NaN into null so "missing last" holds in both directions.
         key = pl.when(key.is_nan()).then(None).otherwise(key)
-    data = data.with_columns(key.alias(_SORT_KEY))
-    tiebreaks = _sortable_tiebreaks(data.drop(_SORT_KEY), sort_by)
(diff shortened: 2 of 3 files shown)


awwesomeman (commented):
Verified locally in a detached worktree at 5597e375 with `uv sync --all-extras`: `ruff check` clean, `pytest tests/test_compare.py` → 43 passed, and the two new tests would fail on main (the `__factrix_sort_key` params column is dropped there; the list-valued permutation sweep is order-dependent there). The internal-name derivation and the canonical key are proportionate to #1092 / #1093.

**Required**

1. `ruff format` fails on `factrix/_compare.py:255` (CI `lint`).

2. `mypy factrix` fails (CI `lint` stops at the format hook, so this has not surfaced there yet):
   ```
   factrix\_compare.py:280: error: List comprehension has incompatible type List[str]; expected List[tuple[str, str]]  [misc]
   factrix\_compare.py:285: error: Generator has incompatible item type "str"; expected "tuple[str, str]"  [misc]
   ```
   `items` in `_canonical_sort_value` is first bound to a `list[tuple[str, str]]` in the `Mapping` branch and then re-bound to `list[str]` in the sequence and set branches. One name per branch fixes it.

**Optional**

- The canonical key is a type-tagged `repr`, so `sort_by` on a list column orders lexicographically on that string rather than element-wise numerically. Measured: three rows with `window` = `[10, 2]`, `[2, 1]`, `[3, 4]` sort as `[[10, 2], [2, 1], [3, 4]]`. That is deterministic, which is what #1093 asks for, and the docs already say the key is a canonical string; one sentence in the "Ties, order, and missing values" section saying the order of nested values is not numeric would save a reader the surprise. Not blocking.

PR #1111, opened 2026-09-07
fix(stats): bound automatic block fallbacks (#1101)

### LINKED ISSUE #1101: fix(stats): the auto block length escapes its own bound, making the documented degenerate fallback unreachable at small n
## Problem

`_politis_white_block_length` (`factrix/_stats/bootstrap.py`) clamps its result to `_max_block_length(n)` on the main path (line 303) but not on the `fallback = max(1.0, 1.75 * n ** (1/3))` early returns (line 210), which are taken when `n < 4` or `gamma_0 < EPSILON` — a constant series. `_block_bootstrap_diff_p` then rejects the value its own resolver produced.

Measured on a constant series through `STATIONARY_BOOTSTRAP.compute` with `block_length` left at its `"auto"` default:

```
n=  2 UserInputError: a block length in [1, 1] for a series of 2 observations
n=  4 UserInputError: a block length in [1, 2] for a series of 4 observations
n=  8 UserInputError: a block length in [1, 3] for a series of 8 observations
n=  9 UserInputError: a block length in [1, 3] for a series of 9 observations
n= 12 UserInputError: a block length in [1, 4] for a series of 12 observations
n= 13 OK  studentized=False  warns=['degenerate_variance','unreliable_se_short_periods']
n= 30 OK  studentized=False  warns=['degenerate_variance']
```

Two things are wrong. The refusal names `block_length`, a field this caller never set, so it points at a knob the user cannot see. And the contract in the `StationaryBootstrap` docstring — "A sample with no usable dispersion admits no such SE, and the kernel falls back to the raw-mean root; that switch is reported as `metadata['studentized'] = False` **and** as `WarningCode.DEGENERATE_VARIANCE`, never silently" — is unreachable for exactly that sample at `n` in {2..9, 12}. At `n >= 13` the documented path works as written.

This is a validation and message defect, not a false inference claim: nothing wrong is published, the call simply fails where the docstring says it degrades.

## Expected

A resolver that produces a block length produces one inside the bound its own validator enforces, and a refusal names a quantity the caller supplied.

## Definition of Done

- [ ] The fallback path clamps to `_max_block_length(n)` like the main path.
- [ ] The degenerate-variance fallback is reachable at every `n` the kernel accepts, so the docstring's "never silently" contract holds where it claims to.
- [ ] A refusal triggered by a resolved default names the resolved default as such, rather than a knob the caller left unset.
- [ ] Tests cover a constant series across the small-`n` range above.

Refs #983

awwesomeman:
## Summary

- cap every Politis-White fallback at the same admissible block-length maximum as the plug-in path
- keep automatic resolution inside the downstream kernel's accepted domain for short, constant, and non-finite series
- pin the small-sample degenerate-variance path so it reaches the documented raw-mean root instead of rejecting an internal default

## Contract

An automatically resolved block length is always inside `[1, ceil(min(3*sqrt(n), n/3))]`. Explicit user-supplied out-of-range values remain validation errors and are not silently clamped.

## Verification

Actually run on this machine:

- `git diff --check` — passed

Actually run by remote CI on current head `b5e7b107`:

- lint and mypy — passed
- pytest on Python 3.12–3.14 across Ubuntu, macOS, and Windows — passed
- documented dev setup and dependency-floor/no-pandas cells — passed
- doctests and wheel import smoke — passed
- 17 successful checks, no failures or pending checks

Closes #1101

### DIFF
--- a/factrix/_stats/bootstrap.py
+++ b/factrix/_stats/bootstrap.py
@@ -202,12 +202,18 @@ def _politis_white_block_length(values: np.ndarray) -> float:
     Falls back to ``max(1, 1.75 · T^(1/3))`` (the widely-cited practical
     PW approximation, also used by ``factrix.stats.bootstrap``) when
     the series is too short, autocovariance is degenerate, or the
-    spectral estimate yields a non-finite ratio. Returns a ``float``;
-    callers that need an integer block size round at the call site.
+    spectral estimate yields a non-finite ratio. The fallback is capped at
+    :func:`_max_block_length`, exactly like the plug-in path, so an automatic
+    resolver never returns a value its own kernel rejects. Returns a
+    ``float``; callers that need an integer block size round at the call site.
     """
     x = np.asarray(values, dtype=float)
     n = len(x)
-    fallback = max(1.0, 1.75 * n ** (1.0 / 3.0)) if n >= 1 else 1.0
+    fallback = (
+        min(max(1.0, 1.75 * n ** (1.0 / 3.0)), float(_max_block_length(n)))
+        if n >= 1
+        else 1.0
+    )
     if n < 4 or not np.all(np.isfinite(x)):
         return fallback
 
--- a/tests/inference/test_series_mean.py
+++ b/tests/inference/test_series_mean.py
@@ -331,6 +331,14 @@ class TestStationaryBootstrap:
         )
         assert WarningCode.UNRELIABLE_SE_SHORT_PERIODS in result.warnings
 
+    @pytest.mark.parametrize("n", [2, 4, 8, 9, 12])
+    def test_constant_small_samples_reach_the_degenerate_fallback(self, n) -> None:
+        result = STATIONARY_BOOTSTRAP.compute(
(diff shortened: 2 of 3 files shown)


awwesomeman (commented):
Verified locally in a detached worktree at a42481e3 with `uv sync --all-extras`: `ruff format --check` clean, `mypy factrix` clean, `pytest tests/stats/test_block_bootstrap.py tests/inference/test_series_mean.py` → 122 passed. The cap is the same `_max_block_length(n)` the plug-in path applies at `bootstrap.py:303`, and the new parametrised bound test would fail on main at n ∈ {2, 4, 8, 9, 12} (fallback 2.2 / 2.8 / 3.5 / 3.6 / 4.0 against b_max 1 / 2 / 3 / 3 / 4 — the n=12 cell passes both ways).

**Required**

1. `ruff check` fails (CI `lint`): `tests/stats/test_block_bootstrap.py:45` — the repo rule wants `pytest.approx(expected) == L`, as the surrounding tests in the same class already write it. `ruff check --fix` resolves it.

Approve once that is in.

PR #1113, opened 2026-09-08
fix(warnings): align warning messages and declaration surfaces

### LINKED ISSUE #1105: fix(metrics): common_beta's few-assets warning names a df the test did not run at
## Problem

`factrix/metrics/common_beta.py:348` — the `FEW_ASSETS` warning tells the reader the cross-asset t-test runs on `df = n - 1`, but the test ran on the Welch-Satterthwaite df stamped one line earlier at `common_beta.py:331`. `n - 1` is the df of the **iid fallback branch only**.

Measured on a 4-asset panel through `compute_common_betas`, with `calendar_time_se_applied=True`:

```
warning: "n_assets=4 ...; the cross-asset t-test on the mean per-asset beta runs on
          df=3, so its critical value is well above the asymptotic one."

metadata["dof"] = 73
reported p_value  = 0.026366     (t = 2.2668)
p at df=3, the df the warning names = 0.108245
```

Two rules break at once. The message reports a quantity the code did not use, and it contradicts the `dof` metadata emitted beside it — CLAUDE.md's "a message naming a metadata key must print the value under that key" and "warning text may not contradict the metadata emitted beside it". The gap is not cosmetic: 0.026 and 0.108 sit on opposite sides of the 5% line, so a reader who trusts the sentence reaches the opposite conclusion about the same estimate.

## Expected

The warning names the df the test actually ran at, or says explicitly which branch its `n - 1` describes and that the calendar-time branch uses a different one.

## Definition of Done

- [ ] The message reads the same quantity it names — `metadata["dof"]` on the calendar-time branch, `n - 1` on the iid fallback.
- [ ] If the screen is deliberately calibrated on the smaller df, the message says both numbers, per the "report what ran" exception rule.
- [ ] A test pins the message against the metadata on both branches, and fails if they diverge again.
- [ ] Sweep the remaining warnings that quote a df or a critical value for the same pattern.

Found by a "report what ran" sweep (52 `_emit_warning` sites, 9 `warnings.warn`, every f-string `raise`) while auditing the closed issues of the #1062 series.

Refs #1035 #1039

### LINKED ISSUE #1106: fix(warnings): the expected_warnings contract is not the universal statement #859 made it
## Problem

#859 set out to make `expected_warnings` a universal statement: every user-visible warning carries a `WarningCode` and can be declared. A mechanical sweep finds three ways out of that contract. All 43 `WarningCode` members are referenced (0 orphans), so the gaps are on the channel side, not the vocabulary side.

### 1. Emitted, not declarable

Two user-visible `UserWarning`s carry no code and have no declaration path:

- `factrix/stats/multiple_testing.py:350` — `romano_wolf_adjusted_p`'s uncentred-bootstrap advisory
- `factrix/stats/multiple_testing.py:468` — its `B < 999` resolution advisory

`spanning.greedy_forward_selection`'s snooping note is the third code-less `UserWarning`, but it has its own `suppress_snooping_warning=` opt-out and #859 recorded that exemption deliberately. These two were never mentioned in either issue.

(Five `RuntimeWarning`s in `_multi_factor.py` — 970, 1218, 1501, 1927, 2013 — are a different category and are out of scope unless the contract is meant to cover them too; say so either way.)

### 2. Recorded, never emitted

Three codes reach `EvaluationResult.warnings` through no channel a reader can see:

```
monotonicity() on 8 assets:  recorded {few_assets, metric_unavailable, thin_quantile_groups}
                             echoed   {few_assets, thin_quantile_groups}
bmp_z() on a dense panel:    recorded {structure_mismatch}          echoed {}
bmp_z() on a 30%-zero panel: recorded {..., frequent_event_signal}  echoed {...}
```

Sources: `_dag.py:424` (`METRIC_UNAVAILABLE` / `UPSTREAM_UNAVAILABLE`), `__init__.py:1024` (`STRUCTURE_MISMATCH`), `__init__.py:1252` (`FREQUENT_EVENT_SIGNAL`). On #925's own wording — "every echo that affects result interpretation" — a metric that silently returned NaN because it was unavailable is exactly such a case.

### 3. `inspect_data` has no declaration surface at all

#925's opening sentence is unaddressed: its merge never touched `_inspect.py`, and `evaluate` does not call `inspect_data`. Measured on an 8-asset, 90-period panel:

```
fx.inspect_data(panel, factor_cols=["factor"])
  echoes: 0
  recorded warnings: 10
  signature: ['data', 'factor_cols']        # no expected_warnings
```

The pre-flight `Warning` objects live on the result only, and the function cannot be told about them.

## Expected

Every user-visible warning has a code and can be declared, and every recorded code that affects interpretation reaches the reader through the one channel.

## Definition of Done

- [ ] The two `multiple_testing` advisories get codes, or their exemption is recorded where #859 recorded `greedy_forward_selection`'s.
- [ ] The three record-only codes either echo or are documented as deliberately silent, with the reason.
- [ ] `inspect_data` accepts `expected_warnings` and routes its pre-flight warnings through the same channel, or the divergence is documented on the function.
- [ ] `tests/test_warning_formatter.py`'s AST guard gains a converse check: today it verifies no code echoes outside the chokepoint, and cannot see a code that never echoes at all.
- [ ] State whether `RuntimeWarning`s are inside the contract.

## Also worth fixing while there

That AST guard decides a site is code-bearing by looking for the literal string `WarningCode` within the 25 lines above the call. A future site appending a code-valued string literal without naming the enum would pass it.

Found by a warning-channel sweep (43 codes, 75 `_emit_warning` references, 9 real `warnings.warn` calls) while auditing the closed issues of the #1062 series.

Refs #859 #925 #1012

awwesomeman:
## Summary

- report the degrees of freedom from the inference branch that actually produced the common-beta warning
- route Romano-Wolf advisories, evaluation-assembly warnings, and inspect-data warnings through the structured warning declaration contract
- tighten static warning-site coverage so raw emissions and structured records remain explicitly inventoried
- document the newly coded warnings and the supported suppression boundaries
- keep executable documentation and the curated LLM warning inventory synchronized with the public contract

## Verification

Actually run on this machine:

- `git diff --check` after each correction — passed
- warning-table row-count check — passed; 37 declared evaluation-side codes and 37 table rows
- commit/trailer inspection — passed; no signature or `Signed-off-by` trailer

Actually run by remote CI on current head `f77ded4d`:

- lint and mypy — passed
- pytest on Python 3.12–3.14 across Ubuntu, macOS, and Windows — passed
- documented dev setup and dependency-floor/no-pandas cells — passed
- executable docs examples, curated LLM inventory tests, and doctests — passed
- documentation build and wheel import smoke — passed
- 18 successful checks, no failures or pending checks

Earlier remote revisions exposed the missing `factrix`/panel setup in the new inspect-data example and the two missing Romano-Wolf rows in `llms-full.txt`; `a2fe6773` and `f77ded4d` corrected those before the successful run above.

Closes #1105
Closes #1106


awwesomeman (commented):
Verified on a local merge of this branch onto current main (2e0de2db, after #1109/#1110/#1111/#1112/#1118 landed; `git merge-tree` clean): `ruff format --check` / `ruff check` clean, `mypy factrix` clean (83 files), full suite 4280 passed, 44 skipped, 0 failed.

I also measured the echo counts the PR changes, per code, on the same panels the new tests use (main vs main + this branch):

| call | main | with #1113 |
|---|---|---|
| `evaluate(spanning_alpha)` unavailable | `thin_quantile_groups` ×1 | + `metric_unavailable` ×1 |
| `evaluate(caar)` on a dense panel | none | `structure_mismatch` ×1 |
| `evaluate(caar)` on a 40%-zero panel | 2 event codes | + `frequent_event_signal` ×1 |
| `evaluate(monotonicity)` on 8 assets | 2 codes | + `metric_unavailable` ×1 |
| `inspect_data` on 8 assets | none | `few_assets` ×9 |

Every previously record-only code now echoes exactly once — the `_dag.py` dedup does its job, no double echo. Both #1105 branches are pinned against the metadata they report (`df=3` iid, `df=73` calendar-time), the two Romano-Wolf advisories carry codes and a declaration path, `inspect_data` accepts `expected_warnings`, the AST guard no longer relies on a 25-line window, and the `RuntimeWarning` boundary is stated in the observability guide. That is every #1106 Definition of Done item.

No required changes.

**Optional**

- `inspect_data` on the 8-asset panel prints nine `few_assets` lines to stderr, one per metric verdict, for one regime. That is the contract as written (each record echoes), but a pre-flight that fans one thin cross-section out into nine identical warnings is loud in the one place users go to *avoid* noise. If it turns out to bother people, echoing per distinct `(code, source-group)` while keeping every record would be a small follow-up; not for this PR.
- `test_warning_record_sites_are_routed_to_an_echo_surface` pins the inventory of `Warning(` call sites; the routing itself is asserted by the comment above it, not measured. The pin is still a real guard (a new site fails it), but the name overclaims slightly. Cosmetic.

PR #1117, opened 2026-09-08
fix(inspect): align common-factor content gates (#1115)

### LINKED ISSUE #1115: fix(inspect): pre-flight marks common_quantile_spread usable on a zero-variance common factor that evaluate refuses

awwesomeman:
## Summary

- pre-flight `common_quantile_spread` on the same finite per-date series and distinct-factor gate as runtime
- apply its period floor to that effective series rather than to raw calendar dates that runtime discards
- re-enable the full public metric inventory for all zero-variance COMMON agreement cells
- follow up merged #1118 so positive and negative one-sided signals describe the correct realised-sign rate
- ignore NaN/inf in pre-flight sign support exactly as runtime does
- correct remaining user docs that still described the finite degenerate hit rate as `value=NaN`

## Quantitative contract

For the default `common_quantile_spread(n_groups=5)`, pre-flight now reports a blocker when the usable per-date history has `n_distinct < n_groups * 2`, including the observed count and bound. The same stage-one profile supplies the effective period count used by the metric's hard floor.

For `directional_hit_rate`, a one-sided factor remains usable with `DEGENERATE_VARIANCE`: the descriptive hit rate survives while `stat` and `p_value` are withheld. On an all-positive factor it is the positive-return rate; on an all-negative factor it is the negative-return rate. The advisory now says the general rule — returns matching the sole predicted sign — and its detector excludes non-finite cells like runtime.

The inspect/evaluate regression grid contains 90 cells per invariant: 30 individual, 30 time-varying COMMON controls, and 30 zero-variance COMMON regressions. Every shape again exercises the same full public metric inventory; the temporary common-beta-only narrowing merged in #1112 is removed.

## Verification

Actually run on this machine for current head `e07d4909`:

- `git diff --check` and staged diff check — passed
- merge-conflict scan — passed; no unmerged entries remain

Not run on this machine:

- Python quality gates — this machine uses the restricted environment described in the repository workflow; no unavailable gate is reported as passed

Remote CI on current head `e07d4909`:

- ruff check and format — passed; mypy found no issues in 83 source files
- full Python 3.12 Ubuntu run — 4277 passed, 44 skipped
- Python 3.12–3.14 across Ubuntu, macOS, and Windows — passed
- documented `--extra dev` setup — 4196 passed, 49 skipped
- dependency-floor/no-pandas cells — passed on Ubuntu, macOS, and Windows
- no-theme docs cell — 21 collected, 19 passed, exactly 2 expected skips
- doctests, strict documentation build, link check, and wheel import smoke — passed
- 19 successful checks, 0 failures, 0 pending; 2 expected workflow skips

The prior head `2d986830` had `ruff check` pass but `ruff format --check` fail on three layout-only changes. Commit `e07d4909` applied the exact formatter diff before the successful run above.

Measured earlier on the separate verification machine for prior head `39ae3195` (before synchronising current `main` and adding the follow-ups above):

- dedicated #1115 tests: 3 passed after the fix; 2 failed and the varying-factor control passed against unfixed `_inspect.py`
- full pytest: 4180 passed, 44 skipped
- ruff check / format, mypy (83 source files), and strict docs build: passed

Measured during #1112 review before #1115/#1116 were fixed: removing the temporary narrowing made all 60 zero-variance agreement cells fail, and `common_quantile_spread` plus `directional_hit_rate` were the only disagreeing metrics. No current-head result is inferred from either earlier measurement.

Closes #1115

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -95,6 +95,57 @@ jobs:
       - name: Run pytest
         run: uv run --python 3.12 pytest -rs
 
+  docs-no-theme:
+    name: pytest (docs machinery, no theme)
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - uses: actions/checkout@v7
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v7
+        with:
+          version: "0.8.x"
+          enable-cache: true
+
+      # This intentionally constructs the one dependency boundary no declared
+      # extra represents: mkdocstrings is importable but mkdocs-material is
+      # absent. It proves the false direction of ``requires_theme``; the
+      # all-extras matrix above proves the two marked tests run with the theme.
+      - name: Install docs machinery without the theme
+        shell: bash
+        run: |
+          uv sync --frozen --python 3.12 --extra dev
+          uv pip install --python .venv \
+            "mkdocstrings==1.0.4" "mkdocstrings-python==2.0.3"
+          .venv/bin/python - <<'PY'
+          import importlib.metadata
+          import importlib.util
+
+          assert importlib.util.find_spec("mkdocstrings") is not None
+          try:
+              importlib.metadata.version("mkdocs-material")
+          except importlib.metadata.PackageNotFoundError:
+              pass
+          else:
+              raise AssertionError("mkdocs-material must be absent in this CI cell")
+          PY
+
+      - name: Assert the requires_theme skips
+        shell: bash
+        run: |
+          set -euo pipefail
+          set +e
+          output=$(uv run --python 3.12 pytest -q -rs --color=no \
+            tests/test_docs_inventory_cache_warmup.py 2>&1)
+          status=$?
+          set -e
+          printf '%s\n' "$output"
+          test "$status" -eq 0
+          grep -Fq "19 passed, 2 skipped" <<<"$output"
+          reason="resolving mkdocs.yml requires mkdocs-material from the docs extra"
+          grep -Fq "$reason" <<<"$output"
+
   floor:
     name: pytest (declared dependency floor, no pandas, ${{ matrix.os }})
     runs-on: ${{ matrix.os }}
--- a/docs/api/compare.md
+++ b/docs/api/compare.md
@@ -57,7 +57,7 @@ The returned `pl.DataFrame` contains the following columns:
 | `metrics` | (required) | `list[str]` of metric labels to include in the leaderboard. |
 | `sort_by` | `None` | Any output column produced before ranking: `factor`, `forward_periods`, `params` keys, metric value columns such as `ic`, or p-value columns such as `ic_p_value` / `<metric_label>_p_value`. `None` keeps the original list order. |
 | `descending` | `None` | Sort direction for `sort_by`. `None` resolves it from the column under the [direction rule](#sort-direction); `True` / `False` states it outright and always wins. There is no global direction default. |
-| `rank_method` | `"min"` | How equal `sort_by` values are numbered: `"min"` (tied rows share the best rank, next rank skips: `1, 1, 3`), `"dense"` (share the rank, no gap: `1, 1, 2`), `"ordinal"` (every row numbered `1..N`). Polars' `Expr.rank` methods. |
+| `rank_method` | `"min"` | How equal `sort_by` values are numbered: `"min"` (tied rows share the best rank, next rank skips: `1, 1, 3`), `"dense"` (share the rank, no gap: `1, 1, 2`), `"ordinal"` (every row numbered `1..N`, using the deterministic [tie order](#ties-order-and-missing-values)). Polars' `Expr.rank` methods. |
 
 ### Sort direction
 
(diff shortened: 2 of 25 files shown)


awwesomeman (commented):
Verified locally at 4e75292d (main `.venv` with `PYTHONPATH` pointed at a detached worktree): `ruff format --check` / `ruff check` clean, `mypy factrix` clean (83 files), `pytest tests/test_inspect_data.py tests/test_inspect_evaluate_agreement.py` → 251 passed. The new `TestCommonQuantileFactorVariation` run against main's `_inspect.py` fails 2 / passes the varying-factor control, as the PR states. The mirror reads the same `_aggregate_to_per_date` series the metric cuts and the same default `n_groups` the run would use, so the two cannot drift.

Also measured what these fixes buy together: with this branch merged with #1118 and #1112's `test_inspect_evaluate_agreement.py` dropped in with the `_COMMON_BETA_CONSUMERS` narrowing removed, the full sweep passes 214 / 214 — all 60 `common_zero_variance` cells over the complete public metric set. So once #1117, #1118 and #1112 are all in, the narrowing can go in a small follow-up.

**Required**

1. `factrix/_inspect.py:78-83` — the attribute docstring `"""Metrics gated on an optional schema column: metric name to declared column. …"""` belonged to `_REQUIRED_OPTIONAL_COLUMNS`; the new `_HISTORICAL_QUANTILE_METRICS` assignment was inserted between the two, so that docstring now sits under the wrong constant. Move the new constant (and its comment) below the docstring.

On the question raised: leaving `common_asymmetry` alone is right. Its run-time gate is per-sign `n_unique >= 2` on the pooled factor, a different comparison from `n_distinct < n_groups * 2`, and it does not appear in the zero-variance disagreement set, so mirroring it would be a guard for a disagreement that has not been measured.

awwesomeman (commented):
Re-reviewed the rebased head e07d4909 (onto 2e0de2db). The docstring-placement fix is in (39ae3195), and the two commits added after my first pass widen the PR beyond the original #1115 scope — all in the right direction:

- the `_COMMON_BETA_CONSUMERS` narrowing in `tests/test_inspect_evaluate_agreement.py` is removed, so the zero-variance sweep now runs over the full public metric set in CI (this is the follow-up I had asked for after #1112; it can be closed as done here);
- `common_quantile_spread`'s period floor is now pre-flighted on the same aggregated per-date series the run uses (`n_periods` on the stage-1 profile), with a test that pins `n_periods=2 < min_periods=3` against the run's `insufficient_portfolio_periods`;
- `_factor_sign_is_one_sided` filters to finite cells, so a NaN / -inf cell cannot manufacture a second sign (tested);
- the negative-side one-sided panel is swept alongside the positive one, and the advisory text no longer says "positive-return rate" — correct, since the rate is of returns matching the sole predicted sign;
- docs that still said `directional_hit_rate` returns `value=NaN` on a degenerate PT variance are corrected to the actual contract (value kept, `stat` / `p_value` withheld), which #1118 had established at the metric but not everywhere it was described.

Verified locally on this head (main `.venv`, `PYTHONPATH` at a detached worktree): `ruff format --check` / `ruff check` clean, `mypy factrix` clean (83 files), `pytest tests/test_inspect_data.py tests/test_inspect_evaluate_agreement.py tests/test_directional_hit_rate.py` → 338 passed, which includes all 214 agreement cells with no narrowing — matching the 214 / 214 I measured on the local merge of #1117 + #1118 earlier.

No required changes. Approve once the remaining CI cells report (13 green, 0 failed, 6 in progress at the time of writing).

PR #1118, opened 2026-09-08
fix(inspect): advise, not refuse, a one-sided directional signal (#1116)

### LINKED ISSUE #1116: fix(inspect): directional_hit_rate's one-sided blocker promises a short-circuit that run time does not perform

awwesomeman:
## Summary

`inspect_data` marked `directional_hit_rate` **unusable** on a one-sided sign signal with a blocker promising that the metric "would return degenerate_variance at run time". The run does no such refusal: with one non-zero sign, `P*` collapses to the realised up-rate, `var_s` is zero, and the metric returns the hit rate with `stat` / `p_value` null under `WarningCode.DEGENERATE_VARIANCE`. `strict=True` therefore ran the shape pre-flight had refused.

## The design call

The issue offered two ways to make pre-flight and run time agree: make the metric short-circuit, or retire the blocker. This takes the second.

A hit rate on a one-sided signal is a defined number — the unconditional positive-return rate — but not a *directional* one, and factrix's canonical rule for that case is [architecture.md](docs/development/architecture.md): a degenerate dispersion withholds the test "while retaining a valid point estimate". Short-circuiting would delete a value the metric can compute and would be the breaking change of the two. The advisory now says what the number means, so a reader is not left to assume it measures skill.

Recorded in the `directional_hit_rate` docstring (which also claimed a short-circuit that never happened) and in `docs/reference/metric-applicability.md`.

## Contract

One-sided `sign(factor)` → `usable=True` carrying:

```
degenerate_variance: one-sided directional signal: sign(factor) has only one
non-zero side, so P* collapses to the realised up-rate and the
Pesaran-Timmermann variance is zero. The metric returns the hit rate — the
unconditional positive-return rate, not a directional one — with a null
statistic and p-value under degenerate_variance. Center, threshold, or encode a
true two-sided directional signal to obtain a test.
```

## Verification

Measured on a verification machine (project venv, `pytest -p no:randomly`):

- the three new tests in `tests/test_inspect_data.py::TestDirectionalHitRateOneSidedSignal` against the **unfixed** `_inspect.py`: 2 failed (pre-flight refused; the run did not), 1 passed (the two-sided control, which must not gain an advisory)
- after the change: 3 passed; whole file 97 passed
- `test_directional_hit_rate_blocks_one_sided_factor_sign` pinned the old blocker and is rewritten as `test_directional_hit_rate_advises_on_a_one_sided_factor_sign` — the same panel, the new contract
- `python -m pytest -q`: **4180 passed, 44 skipped, 0 failed**
- `ruff check` / `ruff format --check`: clean
- `python -m pytest --doctest-modules factrix/metrics/directional_hit_rate.py`: 2 passed

Closes #1116

### DIFF
--- a/docs/reference/metric-applicability.md
+++ b/docs/reference/metric-applicability.md
@@ -142,7 +142,7 @@ Min sample*. `MIN_*` constants resolve to values in the
 
 | Metric | Sample axis | Min sample |
 |---|---|---|
-| [`directional_hit_rate`][factrix.metrics.directional_hit_rate.directional_hit_rate] | pooled `(date, asset)` signs | non-overlapping obs `≥ MIN_DIRECTIONAL_PAIRS_HARD`; warn if below `MIN_DIRECTIONAL_PAIRS_WARN` |
+| [`directional_hit_rate`][factrix.metrics.directional_hit_rate.directional_hit_rate] | pooled `(date, asset)` signs | non-overlapping obs `≥ MIN_DIRECTIONAL_PAIRS_HARD`; warn if below `MIN_DIRECTIONAL_PAIRS_WARN` ; a one-sided `sign(factor)` is usable with a `degenerate_variance` advisory — the hit rate is kept, the PT test withheld |
 
 ## Sample-size constants
 [](){ #sample-size-constants }
--- a/factrix/_inspect.py
+++ b/factrix/_inspect.py
@@ -1148,12 +1148,26 @@ def _evaluate_applicability(
         )
 
     if spec.name == "directional_hit_rate" and factor_sign_one_sided:
-        blockers.append(
-            "one-sided directional signal: sign(factor) has only one non-zero "
-            "side, so directional_hit_rate would return degenerate_variance "
-            "at run time; center, threshold, "
-            "or encode a true two-sided directional signal before using this "
-            "metric"
+        # Advisory, not a blocker: the run does not short-circuit here. It
+        # returns the hit rate — which on a one-sided signal is the
+        # unconditional realised-direction rate — and withholds the PT test
+        # under DEGENERATE_VARIANCE, so pre-flight reports that same outcome
+        # rather than promising a refusal the run never performs (#1116).
+        warnings.append(
+            Warning(
+                code=WarningCode.DEGENERATE_VARIANCE,
+                source=spec.name,
+                message=(
+                    "one-sided directional signal: sign(factor) has only one "
+                    "non-zero side, so P* collapses to the realised up-rate "
+                    "and the Pesaran-Timmermann variance is zero. The metric "
+                    "returns the hit rate — the unconditional positive-return "
+                    "rate, not a directional one — with a null statistic and "
+                    "p-value under degenerate_variance. Center, threshold, or "
+                    "encode a true two-sided directional signal to obtain a "
+                    "test."
+                ),
+            )
         )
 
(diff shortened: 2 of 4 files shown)


awwesomeman (commented):
Verified locally at fe941a06 (main `.venv` with `PYTHONPATH` pointed at a detached worktree): `ruff format --check` / `ruff check` clean, `mypy factrix` clean (83 files), `pytest tests/test_inspect_data.py tests/test_inspect_evaluate_agreement.py` → 251 passed. Against main's `_inspect.py` the rewritten pin plus the two new positive tests fail (3) and the two-sided control passes (1).

Retiring the blocker is the right branch of the design call: the run already keeps `value` and withholds the test under `degenerate_variance`, so pre-flight now reports the outcome the run produces instead of a refusal it never performs, and no metric contract changes. The docstring correction in `directional_hit_rate.py` closes the same gap on the metric side.

Combined with #1117 and the #1112 sweep with its narrowing removed, the full agreement file passes 214 / 214 (all 60 `common_zero_variance` cells over every public metric), measured on a local merge of the two branches.

On the question raised: the advisory is already on the metric's own surface — the run returns `warning_codes` containing `degenerate_variance` — so pre-flight mirroring it under the same `WarningCode` is the complete picture; nothing further needed.

No required changes. Approve once CI settles.

PR #1121, opened 2026-09-08
fix(results): make reporting exports auditable

### LINKED ISSUE #88: feat: Integration adapter design: MLflow logging and Plotly charts
> **Status (2026-09-04)** — **Not `parked`.** Nothing external gates this: it is unscheduled documentation work, and low priority, which the house convention says to state honestly rather than to file behind a trigger that never fires. The `parked` label was removed for that reason. The chart half overlaps #335 — if both proceed, reconcile rather than writing the charts twice.

## Problem

factrix deliberately does not bundle rendering or ML-tracking layers into its core wheel. The `factrix[charts]` and `factrix[mlflow]` optional extras were removed in v0.8.0 because the corresponding source modules (`factrix.charts`, `factrix.integrations.mlflow`) had already been stripped in v0.5.0, leaving the extras as install-only stubs that pulled in heavy transitive dependencies with no factrix code consuming them.

Downstream users who need these integrations have no documented starting point.

## Expected

A dedicated document (or example notebook) providing recommended adapter patterns, so users can wire factrix outputs into their existing toolchains without guesswork.

## Scope

- **MLflow**: how to log `EvaluationResult` fields as MLflow run metrics / tags — per-metric `MetricResult.value` / `MetricResult.p_value` / `n_obs` **paired with `n_obs_axis`** (the count is only interpretable with its axis: `periods` / `events` / `pairs` / `asset_pairs` / `assets`), the bundle-level `EvaluationResult.n_periods` / `n_pairs` / `n_assets`, plus `EvaluationResult.warnings` and the derived `unexpected_warnings` (tag both — `Warning.expected=True` records are declared study design, not alerts). Suggested experiment-naming convention by axis: note the analysis cell is the triple `EvaluationResult.cell = (FactorScope, FactorDensity, DataStructure)`, so a name built from it should use all three or say why it drops the third.
- **Plotly**: canonical chart types for factrix outputs (IC time series, quantile spread bar, CAAR cumulative return, BHY rejection scatter); how to build them from `MetricResult` / `EvaluationResult` — `EvaluationResult.to_frame(metadata=[...])` exports named estimator metadata as extra columns alongside the fixed schema, and `to_dict()` covers nested values — without re-deriving columns. `to_frame` rows already carry the hypothesis identity (`factor` / `forward_periods` / `params` keys), so results stack across factors with `pl.concat`. Overlaps the recipe table in #335; if both proceed, reconcile rather than writing the charts twice.
- **Design rationale**: why factrix stays visualization- and tracking-agnostic (wheel size, dependency churn, user freedom to swap backends).

## Definition of Done

- [ ] Document or example notebook covering the MLflow adapter pattern.
- [ ] Document or example notebook covering the Plotly chart adapter pattern.
- [ ] Design rationale section explaining the no-bundling policy.

> Note: the result surface has moved repeatedly since this issue was filed. `FactorProfile` / `MetricOutput` / `primary_p` went away at v0.14.0; since then `params` split from `metadata`, the injected horizon became `overlap_periods`, `MetricResult.alternative` was added, and `to_frame()` gained a `metadata=` export. Adapters target `EvaluationResult` / `MetricResult` / `MetricResult.p_value`, but re-verify every field name against the public API immediately before writing snippets rather than trusting any version pinned in this body.

### LINKED ISSUE #335: docs(guides): Plotting Recipes — reference snippets for common factor charts
## Problem

factrix's producers return per-period tables — `compute_ic` and `compute_spread_series` as `dict[str, pl.DataFrame]` keyed by factor column, `compute_group_returns` as a single frame — its metrics return scalar `MetricResult` records (`ic`, `ic_trend`, `caar`, ...), and it stops at inference. Users bridging "metric DataFrame → chart" hit the same friction every time: not "how do I call plotly," but "which column maps to which axis, and how do I reshape for the common factor due-diligence charts."

No canonical reference exists; every user re-derives the same handful of recipes.

## Why recipes, not a plot API

- **Scope** — factrix stops at inference; `factrix.plot` invites theme / tooltip / layout / export requests orthogonal to "is the factor real."
- **Dependency cost** — plotly ~15MB + JS; matplotlib C extensions. Heavy even as `[plot]` extras.
- **Leaky abstraction** — every chart needs different group keys, annotations (OOS split, `n=`, `median=`), regime labels. Narrow per-chart funcs multiply; a god-function with dict config is uglier than plotly express itself.
- **Subjectivity** — bin count, rolling window, regime split, overlay choices — every desk overrides defaults.
- **Bridge, not wrapper** — outputs are already plot-friendly; the real gap is column-to-axis mapping, which a wrapper hides.

## Expected

New guide `docs/guides/plotting-recipes.md`:

- Opens with explicit non-goal: factrix does not ship a plotting module; snippets are reference code to copy.
- Plotly primary; matplotlib only where API divergence is material.
- Each recipe: **purpose → callable + columns → 10–20 line snippet → variation hints**.
- Self-contained via `fx.datasets.make_cs_panel` + `fx.preprocess.compute_forward_return` (no external data), matching the convention sealed by #307 / #319 / #322. Snippets stay on the full period grid (`dates=` omitted, so the stamped `overlap_periods` equals `forward_periods`); a recipe using a coarser evaluation grid must say so, because the stamp is then derived from the grid spacing.
- No factrix-specific plot helper; reshape with plain Polars / pandas.
- Linked from `docs/guides/index.md`.

## Scope (pending discussion — recipe list not finalized)

Producer frames come from `compute_ic` / `compute_spread_series` / `compute_group_returns`. Only `compute_ic` is on `factrix.metrics.__all__`; `compute_spread_series` / `compute_group_returns` are rendered API members on `docs/api/metrics/quantile.md` but are not exported from the package namespace. Decide in the thread whether this guide reconciles that split or routes through `evaluate` instead. `compute_spread_series` / `compute_group_returns` take the horizon as `overlap_periods` (periods on the evaluation grid), not `forward_periods`; `compute_ic` takes no horizon parameter.

Initial straw-man for discussion; final list to be agreed in comments before implementation. Adds / drops / merges expected.

| Recipe | factrix callables |
|---|---|
| Cumulative IC (IS/OOS) + IC distribution | `ic` / `compute_ic` |
| Quantile spread time series + mean return by group | `compute_spread_series`, `compute_group_returns` |
| Multi-horizon IC + IC by regime + rolling IC | `ic` swept / grouped, `ic_trend` |
| Cumulative IC: raw vs orthogonalized | `ic` ± `fx.preprocess.orthogonalize_factor` (returns `OrthogonalizeResult`; unwrap the residualized frame before charting) |
| Cumulative IC + quantile returns by universe | per-universe panel partitioning |

Open questions for the discussion thread:

- Which recipes are core vs nice-to-have?
- Are there factor due-diligence charts missing from this list?
- Should event-driven metrics (`caar`, `event_around_return`) get their own recipe section, or is this issue cross-sectional-only?

## Definition of Done

- [ ] Recipe list finalized in comments before drafting starts.
- [ ] `docs/guides/plotting-recipes.md` exists with the agreed recipes, non-goal statement, and four-part structure per recipe.
- [ ] Code blocks self-contained; no external data; no factrix-specific plot helper introduced.
- [ ] Linked from `docs/guides/index.md`.
- [ ] `uv run mkdocs build --strict` passes; `mkdocs serve` spot-check clean.

## Out of scope

`factrix.plot` / `factrix[plot]` extras; `factrix.utils.plot_helpers` (defer until ≥3 recipes need identical reshape); dashboarding (Streamlit / Panel / Dash).

Refs #329

### LINKED ISSUE #352: docs(examples): add real-anomaly replication recipe
## Problem

The `examples/` section ships two synthetic-data recipes (`stock_factor_evaluation.md`, `multi_factor_screening.md`). Both go well beyond Quickstart's 30-second smoke test — 333 lines / 10 sections and 244 lines / 6 sections respectively, against Quickstart's 108 — but both run on `make_cs_panel` output. A first-day quant looking for the natural payoff — "run a real anomaly through the pipeline end-to-end, read the result, draw a conclusion" — has no recipe to follow.

## Expected

A new example page replicates a published anomaly (e.g. momentum, value) end-to-end:

- raw panel → feature construction
- `evaluate(data, metrics={...}, factor_cols=[...])` → `dict[str, EvaluationResult]` keyed by factor column, in `factor_cols` insertion order (per-metric significance read off `MetricResult.p_value`); `metrics` / `factor_cols` are keyword-only
- `multi_factor.bhy(list(results.values()), metrics=[...], q=0.05)` → `dict[str, BhyResult]`; read `survivors` plus the adjusted p-values `adj_p` / `adj_p_all` (`q` is the nominal FDR target input, not an output)
- chart the per-period IC / λ series (consumes plotting recipes from #335)
- interpret the output against the literature claim

## Definition of Done

- [ ] New page under `docs/examples/` running the full pipeline on a real anomaly
- [ ] Cross-link from `examples/index.md` overview
- [ ] Cross-link from `docs/getting-started/quickstart.md` "Next steps" table if the recipe fits as the natural payoff
- [ ] Plot blocks use the plotting recipes from #335
- [ ] `uv run mkdocs build --strict` green; `validation.links` warn count not increased

Blocked by #335.

Refs #336

### LINKED ISSUE #1119: fix(results): preserve hypothesis identity in screening exports
## Problem

Screening result containers define hypothesis identity as `(factor, forward_periods, *params)`, but their `to_frame()` exports retain only `factor` (and, for cross-metric screens, `metric`). A valid family may contain the same factor across horizons or parameter sweeps; those rows then become indistinguishable in the exported frame.

Joining that frame to `EvaluationResult.to_frame()` for charts, experiment tracking, or CSV audit can create ambiguous or Cartesian matches and associate an adjusted p-value with the wrong hypothesis. This affects the shared per-metric FDR result base, `CrossMetricBhyResult`, and `CrossMetricPartialConjunctionResult`.

## Expected

Every screening `to_frame()` preserves the same visible hypothesis identity used by family resolution: `factor`, `forward_periods`, and the union of `EvaluationResult.params` keys in deterministic order. Screening-specific columns follow that identity block. Missing parameter keys remain null rather than dropping rows.

Parameter names that collide with fixed screening columns fail loudly instead of shadowing adjusted p-values or status fields.

## Definition of Done

- [ ] Per-metric BHY, hierarchical BHY, and partial-conjunction exports preserve `forward_periods` and `params`.
- [ ] Cross-metric BHY and cross-metric partial-conjunction exports preserve the same identity before their metric/test columns.
- [ ] Mixed horizons and parameter sweeps remain one row per submitted hypothesis with unambiguous identity.
- [ ] Parameter/fixed-column collisions raise a structured user-input error.
- [ ] API documentation and regression tests state the complete export schema.

Refs #88 #335

### LINKED ISSUE #1120: docs(event): align pre-event direction contract with implementation
## Problem

The shipped implementation and regression contract direction-adjust pre-event event returns:

- `compute_event_returns` computes `sign(factor) * single_bar_return` for `k <= 0`.
- `tests/test_event_horizon.py::test_pre_event_signed` explicitly verifies that positive and negative events do not cancel.

Two public references state the opposite:

- `docs/api/metrics/event_horizon.md` labels the pre-event branch "Sign-adjusted: No".
- `docs/reference/metric-applicability.md` calls it an "unsigned single-bar return".

The real asymmetry is cumulative post-event versus single-bar pre-event, not signed versus unsigned. A reader following the current docs can reverse the interpretation of directional leakage.

## Expected

Document both branches as direction-adjusted, while retaining the deliberate horizon asymmetry:

- `k > 0`: signed cumulative holding return from the common post-event entry.
- `k <= 0`: signed single-bar return at that offset.

The plotting recipe must consume `event_around_return.metadata["per_offset"]` without re-signing or re-accumulating either side.

## Definition of Done

- [ ] API offset table agrees with the implementation and regression test.
- [ ] Metric-applicability reference agrees with the implementation.
- [ ] The event plotting recipe states the cumulative-post / single-bar-pre asymmetry and does not transform the published path.

Refs #335 #88

awwesomeman:
## Summary

- preserve `factor`, `forward_periods`, and deterministic `params` columns in every screening export; partial conjunction omits only components it has already aggregated as conditions
- reject parameter names that shadow fixed screening columns instead of silently associating adjusted p-values with the wrong hypothesis
- add copy-ready Plotly recipes for IC, quantile, event-path, and FDR audit charts plus an MLflow adapter that retains cell identity, sample axes, warnings, and the nested result artifact
- add an executable notebook recipe over the Kenneth R. French Data Library's real 25 Size x Book-to-Market portfolio returns, using the original 1963-07 through 1990-12 study window, a declared two-factor family, BHY, and per-period IC paths
- align the event-horizon docs with the tested implementation: both pre-event and event-day single-bar returns are direction-adjusted; the pre/post asymmetry is single-bar versus cumulative
- keep Plotly and MLflow caller-installed; no package dependency or factrix plotting wrapper is added

## Quantitative contract

A screening row now exposes the same hypothesis identity used by family resolution. For partial conjunction, `expand_over` components are conditions, so exporting one representative condition would be false precision; those components alone are omitted. This includes `forward_periods` when horizon is the declared condition axis.

The event recipe consumes the published `per_offset` values without re-signing or re-accumulating them. Positive offsets remain cumulative from the common post-event entry; non-positive offsets remain direction-adjusted single bars.

The value example is deliberately a portfolio-level reproduction, not an independent firm-level reconstruction. It downloads rather than redistributes the copyrighted source, validates the complete 5 x 5 header before assigning scores, records the archive SHA-256 and source version line, and discloses that the provider may revise historical returns. Five repeated sort scores make `high_tie_ratio` expected and retained; any other warning remains unexpected. The two tested axes are declared before inspecting p-values.

## Documentation execution boundary

Jupyter code cells tagged `illustrative` render as `python title="Illustrative"`. The value notebook remains executable for a reader, while docs CI syntax-compiles rather than executes its network download and optional Plotly backend. This boundary is visible in the parametrized docs-test inventory.

Generated notebook pages link to the stable repository `examples/` directory while displaying the exact source path. A direct `blob/main/<new file>` URL is invalid until a PR merges and made the PR link gate fail despite a valid source file.

## Verification

Verified at `2ab132cf6ca166c4b0633a7f6c07a3e5cb02dd47`.

Actually run on this machine:

- `git diff --check` / staged `git diff --cached --check` ? passed before each commit
- notebook JSON parse ? passed; 12 cells, 5 code cells, all 5 explicitly tagged `illustrative`
- notebook-to-Markdown renderer-equivalence audit ? exact match for the new page
- MkDocs navigation target existence audit ? passed
- live source-shape audit of the official ZIP ? one expected CSV member; 25 expected portfolio columns; 330 complete months from 1963-07 through 1990-12; no missing-value sentinel in the selected window
- live source snapshot observed ? `This file was created using the 202607 CRSP database.`; SHA-256 `a6522ad4ea890fafb3971938292844e256128de0e2f6929166ff2f0f7d0d573a`
- dependency audit of `pyproject.toml` ? no Plotly or MLflow dependency
- commit trailer/signature audit ? no `Signed-off-by`, bot trailer, or signed commit

Actually run by PR CI and passed at the verified SHA:

- pre-commit Ruff check, Ruff format, and mypy (`lint`)
- pytest on Python 3.12, 3.13, and 3.14 across Ubuntu, macOS, and Windows
- declared dependency floor without pandas across Ubuntu, macOS, and Windows
- documented contributor setup (`--extra dev`)
- docs machinery without theme
- `pytest --doctest-modules`
- wheel build and import smoke
- documentation build and PR link check

Not run locally: Ruff, mypy, pytest, or MkDocs build. This worktree environment is intentionally not provisioned with the project Python gates; the results above are from PR CI, not from this machine.

The optional Plotly and MLflow examples and the network-backed value notebook were syntax-compiled by documentation tests but their third-party/network backends were not executed or rendered in CI. They are explicitly marked illustrative.

## References

- Kenneth R. French Data Library: https://mba.tuck.dartmouth.edu/pages/faculty/ken.french/data_library.html
- 25 Size x Book-to-Market portfolio construction: https://mba.tuck.dartmouth.edu/pages/faculty/ken.french/data_library/tw_5_ports.html
- Fama and French (1992): https://doi.org/10.1111/j.1540-6261.1992.tb04398.x
- Plotly dataframe support: https://plotly.com/python/px-arguments/ and https://plotly.com/python/performance/
- MLflow tracking calls: https://mlflow.org/docs/latest/ml/tracking/tracking-api/ and https://mlflow.org/docs/latest/api_reference/python_api/mlflow.html

Closes #1119
Closes #1120
Closes #88
Closes #335
Closes #352


awwesomeman (commented):
Reviewed at `2ab132cf`. The three CI failures the earlier head carried (`c3f51c90`) are fixed, and I re-verified locally rather than reading them off CI.

**What I measured**

Gates on `2ab132cf` (main `.venv`, `PYTHONPATH` at a detached worktree): `ruff format --check` / `ruff check` clean, `mypy factrix` clean (83 files), full suite 4319 passed, 47 skipped, 0 failed.

Regression force of the #1119 guards — the nine new / changed `to_frame` tests run against `main`'s `factrix/_multi_factor.py` with everything else from this branch:

```
9 failed, 113 passed
```

Every one of the nine fails on the unfixed exporter, so the guards are real rather than pinning. The failures are the right ones: missing `forward_periods`, missing `params` columns, and the absent collision rejection.

On the earlier head I reproduced the three failures exactly (`3 failed, 4301 passed`): seven undeclared python fences importing `plotly` / `mlflow`, two of which also failed on `NameError` for `result` and `mlflow` independent of the missing package, plus five unresolvable `factrix.*` references. `de5d1b02` resolves all three.

**Verified premises**

- #1120's docs correction matches the implementation. `factrix/metrics/_primitives/_event_returns.py:226` computes `signed_return = sign(factor) * (prices[idx + k] / prices[idx + k - 1] - 1)` for the pre-event branch, and `tests/test_event_horizon.py::test_pre_event_signed` pins that a directional pre-event drift does not cancel across long and short events. The old "Sign-adjusted: No" row and the "unsigned single-bar return" phrase were both wrong; the replacements are right.
- Declaring the seven fences `title="Illustrative"` is the repo's established convention, not an ad-hoc dodge — 41 declared fences already exist across 20+ authored pages, and no tested page imports a non-dependency library.
- The `illustrative` Jupyter cell tag propagating into the rendered fence is a genuine mechanism rather than a one-off: documented in `docs/development/documentation.md` and pinned by `test_illustrative_notebook_tag_marks_rendered_fence`.
- #352's four Definition-of-Done items are met: the page exists, `docs/examples/index.md` and the quickstart "Next steps" table both link it, and its IC chart links back to the #335 recipe section rather than restating it.

**Optional**

- `2ab132cf` changes the notebook source link on all three example pages from a `blob/main/<file>` deep link to a `tree/main/examples` directory link. Only the new notebook needed that (its blob URL 404s pre-merge, and `link-check.yml` runs on any PR touching `docs/**`), so two pre-existing pages lose a direct link to buy the third a passing check. `.lycheeignore` is the repo's own mechanism for exactly this and carries a documented one-comment-per-entry convention. Either choice is defensible — the directory link stays checked, an ignore entry stops being checked — but the two unrelated pages regressing is worth a line in the body so it is not read as incidental formatting.
- The `tag_namespace = "factrix"` indirection in the MLflow adapter exists only because `test_page_references_resolve` scans raw page text for `factrix.`-prefixed dotted tokens with no escape hatch, so a legitimate tag key has to be written in a way no user would write it. `docs/guides/observability.md:33` already carries the same workaround (`logging.getLogger("factrix" + ".dag")`). That is the guard's problem, not this PR's — filed as #1122.

No required changes.

